PR #3, opened 2016-11-23
Bugfix/fix datastore build

hochung:
- update path of controllers to make retrofit/spring happy
- update Tests to catch exceptions when property types/entity types/entity set/schema are created.
- disable CustomCSVPopulation to avoid polluting database.


src/test/java/com/kryptnostic/datastore/edm/DatastoreTests.java
```diff
@@ -265,7 +276,7 @@ public void testAddExistingPropertyTypeToEntityType() {
         dms.addPropertyTypesToEntityType(ENTITY_TYPE.getNamespace(), ENTITY_TYPE.getName(), properties);
     }
     
-    @Test(expected=BadRequestException.class)
+    @Test(expected=IllegalArgumentException.class)
```
geekbeast: You shouldn't have the expected method if you are catching and handling this exception below.

src/test/java/com/kryptnostic/datastore/edm/DatastoreTests.java
```diff
@@ -307,7 +328,7 @@ public void testAddPropertyToSchema(){
         	caught = t;
         }
         assertNotNull(caught);
-        assertSame(BadRequestException.class, caught.getClass());
+        assertSame(IllegalArgumentException.class, caught.getClass());
```
geekbeast: This seems odd since you are asserting an exception must be thrown and it must be of type IllegalArgumentException. I would use either the test wide annotation or this check, but not both.

src/test/java/com/kryptnostic/datastore/edm/DatastoreTests.java
hochung: The one below is a different method. I can change the expected method for this one to try-and-catch though for consistency.

src/test/java/com/kryptnostic/datastore/edm/DatastoreTests.java
hochung: It's actually two separate methods here.

src/test/java/com/kryptnostic/datastore/edm/DatastoreTests.java
geekbeast: Sounds like the right approach.

PR #16, opened 2016-12-02
Bugfix/update datastore tests

hochung:
Fix Datastore Tests so that entity sets are properly created; also entity set "Employees" got populated


geekbeast:
@hochung  Should this build be passing?

hochung:
@geekbeast yep, thanks for the reminder. Ran a build just now and succeeded.

PR #25, opened 2017-01-07
Feature/write data

hochung:
- Updated CassandraDataManager table structure, and added write method.
- Added test for CassandraDataManager
- Added NewDataController

### DIFF
--- a/src/main/java/com/kryptnostic/datastore/services/CassandraDataManager.java
+++ b/src/main/java/com/kryptnostic/datastore/services/CassandraDataManager.java
@@ -1,5 +1,9 @@
 package com.kryptnostic.datastore.services;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


src/main/java/com/kryptnostic/datastore/services/CassandraDataManager.java
```diff
@@ -35,95 +49,166 @@
...
+        List<ResultSetFuture> results = entities.stream().map( entity -> {
+            BatchStatement batch = new BatchStatement();
+
+            batch.add( writeIdLookupQuery.bind().setUUID( CommonColumns.SYNCID.cql(), syncId )
+                    .setUUID( CommonColumns.ENTITY_SET_ID.cql(), entitySetId )
+                    .setString( CommonColumns.ENTITYID.cql(), entity.getId() ) );
+
+            SetMultimap<UUID, Object> propertyValues = entity.getPropertyValues();
+
+            propertyValues.entries().stream().filter( entry -> authorizedProperties.contains( entry.getKey() ) )
```
geekbeast: We don't need as much locking here. Just need to make sure we synchronously execute writeIdLookupQuery and then we can do all the individual property's asynchronously. We should wait on all the properties as a whole to finish-- should be pretty with `Futures.allAsList(...)`

PR #27, opened 2017-01-07
shield getAcl endpoint

hochung:


src/main/java/com/kryptnostic/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -82,13 +82,21 @@ public Void updateAcl( @RequestBody AclData req ) {
...
+            throw new ForbiddenException( "Only owner of a securable object can access other users' access rights." );
+        }
+    }
+
+    private boolean isOwner( List<AclKey> aclKeys ) {
+        Set<Principal> principals = Principals.getCurrentPrincipals();
+        return authorizations.checkIfHasPermissions( aclKeys, principals, EnumSet.of( Permission.OWNER ) );
     }
 
     private boolean isOwnerOrCanAlter( List<AclKey> aclKeys ) {
```
geekbeast: Feel free to delete this function as we're merging these permissions.

geekbeast:
I don't think this PR makes sense anymore.

hochung:
We still need to shield the getAcl endpoint, which allow owner of a securable object to get all the permissions related to that object.

geekbeast:
You are correct. What I meant is that I fixed it in a similar way in another PR.

PR #44, opened 2017-01-17
minor fixes of authorizations

hochung:
- bugfix of authorizations controller
- add authorizations mvc pod


src/main/java/com/dataloom/datastore/pods/AuthorizationsMvcPod.java
```diff
@@ -0,0 +1,71 @@
...
+@ComponentScan(
+    basePackageClasses = { AuthorizationsController.class },
+    includeFilters = @ComponentScan.Filter(
+        value = { org.springframework.stereotype.Controller.class,
+                org.springframework.web.bind.annotation.ControllerAdvice.class },
+        type = FilterType.ANNOTATION ) )
+@EnableAsync
+@EnableMetrics(
+    proxyTargetClass = true )
+public class AuthorizationsMvcPod extends WebMvcConfigurationSupport {
```
geekbeast: Please add to DatastoreMvcPod instead of create separate one.

src/main/java/com/dataloom/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -44,7 +44,7 @@
     @RequestMapping(
```
geekbeast: Please ensure your mappings are consistent with what is in other API and controllers.

src/main/java/com/dataloom/datastore/pods/AuthorizationsMvcPod.java
hochung: Is this not considered a separate service? I'm seeing that different services in datastore are getting their own mvc pod right now

src/main/java/com/dataloom/datastore/permissions/controllers/PermissionsController.java
hochung: hm, I think that is the case? Or do you mean I should put the controller path up top rather than setting the absolute path at each endpoint?

src/main/java/com/dataloom/datastore/pods/AuthorizationsMvcPod.java
geekbeast: You should merge in latest changes.

src/main/java/com/dataloom/datastore/permissions/controllers/PermissionsController.java
geekbeast: Exactly!

PR #72, opened 2017-01-22
bugfix/BACKEND-745

hochung:
avoid NPE


src/main/java/com/dataloom/datastore/edm/controllers/EdmController.java
```diff
@@ -143,15 +146,28 @@ private void updateEntitySetIdsToGet(
             Set<UUID> propertyTypeIds,
             Set<UUID> entityTypeIds,
             Set<UUID> entitySetIds ) {
+        boolean setRetrieved = false;
```
geekbeast: code smell.

geekbeast: I haven't had the chance to think about it too deeply, but we should mark this for attention post-Jan 24th.

hochung: sure; just the quickest way I can think of to avoid doing IMap.get whenever possible - probably bad practice.

PR #98, opened 2017-02-07
Feature/index organizations

katherinebernstein:


src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
```diff
@@ -104,4 +104,18 @@ public String executeQueryJson( @RequestBody SearchRequest request ) {
...
     }
+
+    @RequestMapping(
+            path = { ORGANIZATIONS },
+            method = RequestMethod.POST,
+            produces = { MediaType.APPLICATION_JSON_VALUE } )
+    @Override
+    public String executeOrganizationSearch( @RequestBody String searchTerm ) {
+        try {
+            return ObjectMappers.getJsonMapper().writeValueAsString( searchService.executeOrganizationKeywordSearch( searchTerm ) );
```
UnsungHero97: why does this have to be converted to a `String`? what format does `executeOrganizationKeywordSearch()` return?

src/main/java/com/dataloom/organizations/controllers/OrganizationsController.java
```diff
@@ -81,13 +86,14 @@ public Void destroyOrganization( @PathVariable( ID ) UUID organizationId ) {
...
         organizations.destroyOrganization( organizationId );
+        securableObjectTypes.deleteSecurableObjectType( ImmutableList.of( organizationId ) );
         return null;
     }
 
     @Override
     @PutMapping(
         value = ID_PATH + TITLE,
-        consumes = MediaType.TEXT_PLAIN_VALUE )
+        consumes = MediaType.APPLICATION_JSON_VALUE )
```
UnsungHero97: `@RequestBody String title` is still only a `String`, not a JSON object, so does this need to change to `APPLICATION_JSON_VALUE`?

src/main/java/com/dataloom/organizations/controllers/OrganizationsController.java
```diff
@@ -98,7 +104,7 @@ public Void updateTitle( @PathVariable( ID ) UUID organizationId, @RequestBody S
     @Override
     @PutMapping(
         value = ID_PATH + DESCRIPTION,
-        consumes = MediaType.TEXT_PLAIN_VALUE )
+        consumes = MediaType.APPLICATION_JSON_VALUE )
```
UnsungHero97: `@RequestBody String description` is still only a `String`, not a JSON object, so does this need to change to `APPLICATION_JSON_VALUE`?

src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
katherinebernstein: so that we're returning JSON to the frontend--executeOrganizationKeywordSearch returns a List<Map<String, Object>>

src/main/java/com/dataloom/organizations/controllers/OrganizationsController.java
katherinebernstein: for some reason I was getting the title as null with TEXT_PLAIN_VALUE. But also since all the other endpoints take json, it seems cleaner anyway to be consistent.

src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
```diff
@@ -104,4 +104,18 @@ public String executeQueryJson( @RequestBody SearchRequest request ) {
...
+        } catch ( JsonProcessingException e ) {
+            e.printStackTrace();
+        }
+        return "[]";
```
UnsungHero97: I think this should be wrapped in a `finally {}` block

src/main/java/com/dataloom/datastore/services/SearchService.java
```diff
@@ -94,5 +117,35 @@ public void deleteEntitySet( EntitySetDeletedEvent event ) {
...
+    public List<Map<String, Object>> executeOrganizationKeywordSearch( String searchTerm ) {
+        try {
+            List<Map<String, Object>> queryResults = executor.submit( ConductorCall
+                    .wrap( Lambdas.executeOrganizationKeywordSearch( searchTerm ) ) )
+                    .get();
+            return queryResults;
+        } catch ( InterruptedException | ExecutionException e ) {
+            logger.error( "Unable to to perofrm keyword search.", e );
+        }
+        return null;
```
UnsungHero97: is it better to return `null` or an empty list?

UnsungHero97: also, I think this should be wrapped in a`finally {}` block

katherinebernstein: oh yeah good catch thanks, fixed

src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
katherinebernstein: I moved it into the catch (eclipse got mad when I tried to return from a finally). but yeah this is better than having it at the end, thanks

src/main/java/com/dataloom/datastore/services/SearchService.java
katherinebernstein: moved into the catch

src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
geekbeast: So it makes sense to return a string here, but you should be passing through the string returned from elasticsearch. No point in deserializing unless we are going to introspect on contents.

src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
geekbeast: We only use finally blocks if we need to clean something up such as a lock other resource not covered with try with resources.

src/main/java/com/dataloom/organizations/controllers/OrganizationsController.java
geekbeast: This should be TEXT_PLAIN_VALUE. Something must be wrong elsewhere if you are getting a null request body. @UnsungHero97 Could loom-data-js be sending it with the wrong content type header or wrong encoding? Content type header probably correct or being ignored since 400 isn't thrown. Probably bad encoding.

src/main/java/com/dataloom/organizations/controllers/OrganizationsController.java
geekbeast: Same as above.

src/main/java/com/dataloom/organizations/controllers/OrganizationsController.java
geekbeast: Also tagging @hochung since he added a converter to handle strings in Java.

hochung: @katherinebernstein String input should work (i.e. not null) if you pulled `rhizome-client` lately (i.e. last week or so). Let me know if you still have the same issue if you pulled.

katherinebernstein: ah yeah this is working now, thanks!

src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
katherinebernstein: we are introspecting on contents to get permissions

geekbeast: ? If we always have to deserialize then return a List<Map<String, Object>> It didn't seem like we were reading stuff out in this particular.

geekbeast: https://jira.thedataloom.com/browse/BACKEND-807

PR #100, opened 2017-02-08
yaml

katherinebernstein:


src/main/java/com/dataloom/datastore/edm/controllers/EdmController.java
```diff
@@ -196,6 +201,59 @@ public Schema getSchemaContents(
...
+            response.setHeader( "Content-Disposition",
+                    "attachment; filename=" + fileName + "." + fileType.toString() );
+        }
+    }
+    
+    @RequestMapping(
+            path = SCHEMA_PATH + NAMESPACE_PATH + NAME_PATH,
+            method = RequestMethod.GET,
+            produces = { MediaType.APPLICATION_JSON_VALUE, CustomMediaType.TEXT_YAML_VALUE } )
+    public String getSchemaContentsFormatted(
```
geekbeast: Return Schema and just register the YamlMapper as handling the custom media. It will automatically handle selecting the write mapper when serializing the schemas

PR #123, opened 2017-02-24
search Auth0 users by matching an exact email

UnsungHero97:


src/main/java/com/dataloom/datastore/directory/controllers/PrincipalDirectoryController.java
```diff
@@ -120,6 +120,21 @@ public Void removeRoleFromUser( @PathVariable( USER_ID ) String userId, @PathVar
...
+
+    @Override
+    @GetMapping(
+            path = USERS + SEARCH_EMAIL + EMAIL_SEARCH_QUERY_PATH,
+            produces = MediaType.APPLICATION_JSON_VALUE )
+    public Map<String, Auth0UserBasic> searchAllUsersByEmail( @PathVariable( SEARCH_QUERY ) String emailSearchQuery ) {
+
+        // to search by an exact email, the search query must be in this format: email.raw="hristo@kryptnostic.com"
+        // https://auth0.com/docs/api/management/v2/user-search#search-by-email
+        String exactEmailSearchQuery = "email.raw:\"" + emailSearchQuery + "\"";
```
geekbeast: This is definitely the wrong space to build this query.  Did you try using email.raw as a query param?

geekbeast (changes requested):

src/main/java/com/dataloom/datastore/directory/controllers/PrincipalDirectoryController.java
UnsungHero97: Where does this belong?

Here, `exactEmailSearchQuery` * **is** * the query param. Passing this into `userDirectoryService.searchAllUsers()` becomes `GET	/api/v2/users?q=email.raw:"foo@bar.com"`, which is what we want for searching by an exact email:

https://auth0.com/docs/api/management/v2/user-search#search-by-email

geekbeast (approved):

PR #126, opened 2017-02-27
return a 400 not a 500 for an empty search

katherinebernstein:
BACKEND-734

### DIFF
--- a/src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
+++ b/src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
@@ -77,20 +77,14 @@ public class SearchController implements SearchApi, AuthorizingComponent {
         path = { "/", "" },
         method = RequestMethod.POST,
         produces = { MediaType.APPLICATION_JSON_VALUE } )
+    @Override
     public SearchResult executeEntitySetKeywordQuery(
-            @RequestBody Search search,
-            HttpServletResponse response ) throws IOException {
+            @RequestBody Search search ) {
         if ( !search.getOptionalKeyword().isPresent() && !search.getOptionalEntityType().isPresent()
                 && !search.getOptionalPropertyTypes().isPresent() ) {
-            response.sendError( HttpServletResponse.SC_BAD_REQUEST,
+            throw new IllegalArgumentException(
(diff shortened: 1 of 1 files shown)


src/main/java/com/dataloom/datastore/search/controllers/SearchController.java
```diff
@@ -57,31 +59,38 @@
...
+        method = RequestMethod.POST,
+        produces = { MediaType.APPLICATION_JSON_VALUE } )
+    public SearchResult executeEntitySetKeywordQuery(
+            @RequestBody Search search,
+            HttpServletResponse response ) throws IOException {
         if ( !search.getOptionalKeyword().isPresent() && !search.getOptionalEntityType().isPresent()
                 && !search.getOptionalPropertyTypes().isPresent() ) {
-            throw new BadRequestException(
-                    "You must specify at least one request body param (keyword 'kw', entity type id 'eid', or property type ids 'pid'" );
+            response.sendError( HttpServletResponse.SC_BAD_REQUEST,
```
hochung: For consistency we probably want to throw an exception, and return the http response in `DatastoreExceptionHandler` instead.

hochung:
@katherinebernstein  you will need to ask @mindblight for what the original issue is, but I strongly suspect/vaguely remember when he says empty search, he means a search with no result rather than a search with missing params. The current code should be throwing 400 for a missing param search request.

katherinebernstein:
@hochung (github won't let me reply to your comment) --I'm like 95% sure this was about searches with no parameters (I had a few conversations with arion about this, and his concern was making sure requests were formatted properly). Also the exception I had been throwing wasn't hitting DatastoreExceptionHandler, and was returning a 500, so maybe instead we should be adding BadRequestException to the list of exceptions in DatastoreExceptionHandler that throw a 400? I'm not sure which is preferable

hochung:
ah, cool! Somehow I thought that `BadRequestException` is already handled. My poor memory :/ Either way is fine for me (Adding BadRequestException to the handler/Using IllegalArgumentException as it is)

hochung (approved):

katherinebernstein:
haha yeah I thought it would be too--hence why I was originally using an HttpServletResponse lol

PR #130, opened 2017-03-02
Bugfix/backend 841

hochung:
- Fail creation of entity data if user doesn't have write access to all key property types
- `BootstrapDatastoreWithCassandra` can't start up sometimes because of a `SocketTimeOutError` when logging into the four different test accounts. Auth0 Free Tier is imposing a rate limit of [2 API calls per second](https://auth0.com/docs/connections/database/rate-limits), causing the issue. Fixing this by sleeping 2 seconds after each login of test account.


hochung:
looks like there is still a thread interference in bamboo even after I added Thread.sleep :/

geekbeast (changes requested):
This should work better than sleeping threads :) 
https://github.com/kryptnostic/rhizome/pull/188

hochung:
👍 that's cool!

geekbeast (approved):

geekbeast:
@hochung What is missing to merge this?

hochung:
Changes in BACKEND-859, one sec

hochung:
Looks like it's Katherine's branch for TopUtilizers

PR #150, opened 2017-04-05
Feature/graph api refactor

hochung:


src/main/java/com/dataloom/datastore/services/LinkingService.java
```diff
@@ -80,7 +83,7 @@ public UUID link( UUID linkedEntitySetId, Set<Map<UUID, UUID>> linkingProperties
 
         // Warning: We assume that the restrictions on links are enforced/validated as specified in LinkingApi. In
         // particular, from now on we work on the assumption that only identical property types are linked on.
-        initializeComponents( linkIndexedByEntitySets.keySet(), linkIndexedByPropertyTypes, linkIndexedByEntitySets );
+        initializeComponents( dsm.getLatestSyncId( linkIndexedByEntitySets.keySet() ), linkIndexedByPropertyTypes, linkIndexedByEntitySets );
```
katherinebernstein, line 87: this is unnecessary--the search across indices function in SearchService takes the same input, and uses the current sync id for each of the entity sets. So it will still expect Set<UUID> of entity set ids.

katherinebernstein: (meant to leave this on the SearchService function part--I can also update the function though to take in a map instead if you need to generate the map anyway)

hochung: Updating the function to take in a map would be great - the reason is that the Blocker (not just the step in blocker that calls elastic search) needs the syncId (for example to load the data about each entity before performing the search). So it feels more natural for me to specify the SyncId already at this step.

katherinebernstein (approved):

PR #154, opened 2017-04-06
submitAll() should submit all

UnsungHero97:


hochung (approved):

geekbeast (commented):
Did you test this change locally? This was a work around for the submission bug that we were experiencing.

hochung:
@geekbeast that bug was caused by equals method of `Status` object. This `submitAll` was our first suspicion, but I don't think it was the cause.

PR #161, opened 2017-04-19
get all association entity types endpoint

katherinebernstein:


src/main/java/com/dataloom/datastore/edm/controllers/EdmController.java
```diff
@@ -543,6 +543,15 @@ public UUID createEntityType( @RequestBody EntityType entityType ) {
...
         return modelService.getEntityTypes()::iterator;
     }
+    
+    @Override
+    @RequestMapping(
+        path = ASSOCIATION_TYPE_PATH,
+        method = RequestMethod.GET,
+        produces = MediaType.APPLICATION_JSON_VALUE )
+    public Iterable<EntityType> getAssociationEntityTypes() {
+        return modelService.getAssociationEntityTypes()::iterator;
```
geekbeast, line 553: Does this need authorization?

geekbeast (commented):
We should make a task to review authorization on endpoints.

src/main/java/com/dataloom/datastore/edm/controllers/EdmController.java
katherinebernstein: nope, it's just entity types

PR #166, opened 2017-04-25
return empty list if request is empty

katherinebernstein:


geekbeast (approved):
We'll need to do better validation as there's a lot of queries that will end up with zero.

katherinebernstein:
what do you mean a lot of queries will end up with zero?

PR #167, opened 2017-04-26
EntitySetData should have properties ordered

hochung:
Endpoints in `DataGraphManager` for reading entity sets now require an extra param of `orderedPropertyFqns`, which is a `LinkedHashSet` of property types names, in the same order when the entity type is created. 

Accordingly,
- update `loadEntitySetData`, `loadNormalEntitySetData`, `loadLinkedEntitySetData` endpoints in `DataController`
- update `mergeEntities` endpoint in `LinkingService`
- update `DataManagerTest`

### DIFF
--- a/src/main/java/com/dataloom/datastore/data/controllers/DataController.java
+++ b/src/main/java/com/dataloom/datastore/data/controllers/DataController.java
@@ -70,6 +70,7 @@ import com.dataloom.data.EntitySetData;
 import com.dataloom.data.requests.Association;
 import com.dataloom.data.requests.BulkDataCreation;
 import com.dataloom.data.requests.EntitySetSelection;
+import com.dataloom.data.requests.FileType;
 import com.dataloom.datastore.constants.CustomMediaType;
 import com.dataloom.datastore.services.SyncTicketService;
 import com.dataloom.edm.processors.EdmPrimitiveTypeKindGetter;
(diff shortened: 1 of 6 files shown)


src/main/java/com/dataloom/datastore/data/controllers/DataController.java
```diff
@@ -210,21 +237,30 @@ private EntitySetData loadNormalEntitySetData(
                     EnumSet.of( Permission.READ ) );
         }
 
+        LinkedHashSet<FullQualifiedName> orderedPropertyFqns = authzHelper.getAllPropertiesOnEntitySet( entitySetId )
```
katherinebernstein: why does EdmAuthorizationHelper not return a LinkedHashSet<UUID>?

katherinebernstein (approved):

src/main/java/com/dataloom/datastore/data/controllers/DataController.java
hochung: `EntityType.getProperties()` is a `LinkedHashSet` wrapped by `Collection.unmodifiableSet` right now. Not sure if we really want to remove `Collection.unmodifiableSet`, so I didn't change it, but that prevents me from returning `LinkedHashSet`. Since that's just an wrapper, EdmAuthorizationHelper would still return the property Ids in the right order

katherinebernstein: ah makes sense

PR #174, opened 2017-05-06
load search results from cassandra, return DataSearchResult obj

katherinebernstein:


UnsungHero97:
The build is failing because `DatastoreServicesPod` needs `NeuronPod`, which is added in https://github.com/dataloom/datastore/pull/173.

UnsungHero97 (approved):

PR #177, opened 2017-05-09
endpoint for getting available association types for an entity type

katherinebernstein:


geekbeast (approved):

src/main/java/com/dataloom/datastore/edm/controllers/EdmController.java
```diff
@@ -775,10 +775,19 @@ public AssociationType getAssociationTypeById( @PathVariable( ID ) UUID associat
...
     public AssociationDetails getAssociationDetailsForType( @PathVariable( ID ) UUID associationTypeId ) {
         return modelService.getAssociationDetails( associationTypeId );
     }
+
+    @Override
+    @RequestMapping(
+        path = ASSOCIATION_TYPE_PATH + ID_PATH + AVAILABLE_PATH,
+        method = RequestMethod.GET,
+        produces = MediaType.APPLICATION_JSON_VALUE )
+    public Iterable<EntityType> getAvailableAssociationTypesForEntityType( @PathVariable( ID ) UUID entityTypeId ) {
```
UnsungHero97, line 790: The name of this method is confusing. The method is returning `Iterable<EntityType>`, which is a collection of `EntityType`, but the name suggest that it should be returning a collection `AssociationType`.

UnsungHero97 (changes requested):

PR #193, opened 2017-08-03
Refactor + update gitignore

geekbeast:


src/main/java/com/dataloom/datastore/services/SearchService.java
```diff
@@ -520,4 +491,19 @@ private NeighborEntityDetails getNeighborEntityDetails(
...
+            Map<UUID, PropertyType> authorizedPropertyTypes ) {
+        result.getHits().parallelStream()
+                .map( hit -> (String) hit.get( "id" ) )
+                .map( UUID::fromString )
+                .collect( Collectors.toSet() );
+        Set<UUID> ids = result.getHits().parallelStream()
+                .map( hit -> (String) hit.get( "id" ) )
+                .map( UUID::fromString )
+                .collect( Collectors.toSet() );
+        return dataManager.getEntities( ids, authorizedPropertyTypes ).collect( Collectors.toList() );
```
katherinebernstein: so we had been manually adding in the "id" field to each entity before returning it (ex. old version line 265). that is super hacky, but the frontend is currently expecting each entity to have an id field, and will be broken without it. we should probably just have some kind of entity pojo that we return with the id and the entity details, although that will require a bit of frontend modiciation as well as neighbor loading logic (or really anywhere that passes data to the frontend)

geekbeast: The id is added in dataManager.

geekbeast: but yes that is a bit hacky, we should just make every entity have an implicit property called id with reserved FQN.

PR #233, opened 2017-12-09
speed up load entity neighbors

katherinebernstein:


src/main/java/com/dataloom/datastore/services/SearchService.java
```diff
@@ -442,10 +464,10 @@ public SearchResult executeFQNPropertyTypeSearch( String namespace, String name,
 
     }
 
-    private boolean getAuthorization( UUID entitySetId ) {
-        return authorizations.checkIfHasPermissions( new AclKey( entitySetId ),
-                Principals.getCurrentPrincipals(),
-                EnumSet.of( Permission.READ ) );
+    private boolean getAuthorization( UUID entitySetId, Set<Principal> principals ) {
```
geekbeast, line 467: You should implement AuthorizingComponent eventually, we should have an optimized and correct version of authz code that is centralized to make it easier to maintain.

geekbeast (approved):

PR #244, opened 2018-02-09
grant entity set permissions to organization owners when an app type …

katherinebernstein:
…is added to an app


UnsungHero97 (approved):
Unit tests?

PR #257, opened 2018-07-19
filter to authorized roles on getOrganizations

katherinebernstein:


geekbeast (approved):
Should be fine. Monitor for perf changes. There's .distinct() in the stream before you do .parallel()

PR #261, opened 2018-07-21
Feature/background indexing service

katherinebernstein:


src/main/java/com/openlattice/datastore/services/BackgroundIndexingService.java
```diff
@@ -0,0 +1,147 @@
...
+                .valueOf( BLOCK_SIZE )
+                + " OFFSET ?";
+    }
+
+    private Set<UUID> getDirtyEntityKeyIds( UUID entitySetId, long offset ) {
+        String entitySetTableName = DataTables.entityTableName( entitySetId );
+
+        try ( Connection connection = hds.getConnection();
+                PreparedStatement ps = connection.prepareStatement( getDirtyEntitiesQuery( entitySetId ) ) ) {
+            ps.setString( 1, entitySetTableName );
```
geekbeast: This no longer required as you changed query string.

src/main/java/com/openlattice/datastore/services/BackgroundIndexingService.java
```diff
@@ -0,0 +1,147 @@
...
+
+        indexUpdatedEntitySets();
+
+    }
+
+    private String getDirtyEntitiesQuery( UUID entitySetId ) {
+        return "SELECT * FROM " + DataTables.entityTableName( entitySetId ) + " WHERE "
+                + PostgresColumn.LAST_INDEX_FIELD + " < " + PostgresColumn.LAST_WRITE_FIELD + " LIMIT " + String
+                .valueOf( BLOCK_SIZE )
+                + " OFFSET ?";
```
geekbeast: You want to use limit instead of offset. Off set will skip block size rows instead of limiting response to block_size rows.

src/main/java/com/openlattice/datastore/services/BackgroundIndexingService.java
```diff
@@ -0,0 +1,147 @@
...
+        } catch ( SQLException e ) {
+            logger.debug( "Unable to load all entity set ids", e );
+            return ImmutableSet.of();
+        }
+    }
+
+    private Map<UUID, Map<UUID, PropertyType>> getPropertyTypesByEntityTypesById() {
+        Map<UUID, Map<UUID, PropertyType>> result = Maps.newHashMap();
+
+        entityTypes.entrySet().forEach( entry -> {
```
geekbeast, line 102: Would be so much nicer in kotlin :-p

geekbeast (changes requested):

src/main/java/com/openlattice/datastore/services/BackgroundIndexingService.java
```diff
@@ -0,0 +1,147 @@
...
+                    .filter( ptEntry -> !ptEntry.getValue().getDatatype().equals( EdmPrimitiveTypeKind.Binary ) )
+                    .collect(
+                            Collectors.toMap( ptEntry -> ptEntry.getKey(), ptEntry -> ptEntry.getValue() ) );
+            result.put( entityTypeId, propertyTypesById );
+        } );
+
+        return result;
+    }
+
+    @Scheduled( fixedDelay = 500 )
```
geekbeast: Should be @fixedRate You can also specify an initial delay I think

src/main/java/com/openlattice/datastore/services/BackgroundIndexingService.java
```diff
@@ -0,0 +1,147 @@
...
+
+        Map<UUID, Map<UUID, PropertyType>> propertyTypesByEntityType = getPropertyTypesByEntityTypesById();
+
+        entitySets.values().forEach( entitySet -> {
+            if ( !entitySet.getName().equals( AuditEntitySetUtils.AUDIT_ENTITY_SET_NAME ) ) {
+                UUID entitySetId = entitySet.getId();
+                Map<UUID, PropertyType> propertyTypeMap = propertyTypesByEntityType
+                        .get( entitySet.getEntityTypeId() );
+
+                long offset = 0;
```
geekbeast: I see now. You are using offset to implement cursor behavior.  You should actually set fetchSize on the prepared statement and let postgres automatically handle the buffering as you scan out a number of fixed size chunks. 

More details here: https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor

geekbeast (changes requested):

geekbeast (approved):

PR #268, opened 2018-07-27
on full entity set data reindex, don't include audit entity set

katherinebernstein:


geekbeast (commented):
We want to include audit entity set in reindex as it will have data we want to search. Is this causing problems?

katherinebernstein:
yes

PR #279, opened 2018-08-28
a new endpoint for deleting a given entity and all of its neighbor entities

UnsungHero97:
https://github.com/openlattice/api/pull/217
https://github.com/openlattice/conductor-client/pull/320

### DIFF
--- a/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
+++ b/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
@@ -29,6 +29,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.SetMultimap;
 import com.openlattice.authorization.AclKey;
@@ -428,45 +429,79 @@ public class DataController implements DataApi, AuthorizingComponent {
     ) {
 
         ensureReadAccess( new AclKey( entitySetId ) );
+
+        Map<UUID, Map<UUID, PropertyType>> authorizedPropertyTypes = Maps.newHashMap();
+        authorizedPropertyTypes.put(
+                entitySetId,
+                authzHelper.getAuthorizedPropertyTypes( entitySetId, WRITE_PERMISSION )
+        );
+
+        /*
+         * 1 - get the entity neighbors
+         */
+
         Map<UUID, List<NeighborEntityDetails>> result = searchService
                 .executeEntityNeighborSearch( entitySetId, ImmutableSet.of( entityKeyId ) );
 
         if ( result != null && result.containsKey( entityKeyId ) ) {
 
+            /*
+             * 1.1 - collect authorized PropertyTypes per EntitySet
+             */
+
             List<NeighborEntityDetails> neighbors = result.get( entityKeyId );
             neighbors.parallelStream().forEach( neighbor -> {
+                UUID associationEntitySetId = neighbor.getAssociationEntitySet().getId();
+                if ( !authorizedPropertyTypes.containsKey( associationEntitySetId ) ) {
+                    authorizedPropertyTypes.put(
+                            associationEntitySetId,
+                            authzHelper.getAuthorizedPropertyTypes( associationEntitySetId, WRITE_PERMISSION )
+                    );
+                }
+                if ( neighbor.getNeighborEntitySet().isPresent() ) {
+                    UUID neighborEntitySetId = neighbor.getNeighborEntitySet().get().getId();
+                    if ( !authorizedPropertyTypes.containsKey( neighborEntitySetId ) ) {
+                        authorizedPropertyTypes.put(
+                                neighborEntitySetId,
+                                authzHelper.getAuthorizedPropertyTypes( neighborEntitySetId, WRITE_PERMISSION )
+                        );
+                    }
+                }
+            } );
 
+            /*
+             * 1.2 - clear all neighbor and association entities
+             */
+
+            neighbors.parallelStream().forEach( neighbor -> {
                 UUID associationEntityKeyId = (UUID) neighbor.getAssociationDetails().get( ID_FQN ).iterator().next();
                 UUID associationEntitySetId = neighbor.getAssociationEntitySet().getId();
-                Map<UUID, PropertyType> authorizedPropertyTypesOnNeighborAES = authzHelper
-                        .getAuthorizedPropertyTypes( associationEntitySetId, WRITE_PERMISSION );
                 dgm.clearEntities(
-                        neighbor.getAssociationEntitySet().getId(),
+                        associationEntitySetId,
                         ImmutableSet.of( associationEntityKeyId ),
-                        authorizedPropertyTypesOnNeighborAES
+                        authorizedPropertyTypes.get( associationEntitySetId )
                 );
-
                 if ( neighbor.getNeighborDetails().isPresent() && neighbor.getNeighborEntitySet().isPresent() ) {
                     SetMultimap<FullQualifiedName, Object> neighborDetails = neighbor.getNeighborDetails().get();
                     UUID neighborEntityKeyId = (UUID) neighborDetails.get( ID_FQN ).iterator().next();
                     UUID neighborEntitySetId = neighbor.getNeighborEntitySet().get().getId();
-                    Map<UUID, PropertyType> authorizedPropertyTypesOnNeighborES = authzHelper
-                            .getAuthorizedPropertyTypes( neighborEntitySetId, WRITE_PERMISSION );
                     dgm.clearEntities(
                             neighborEntitySetId,
                             ImmutableSet.of( neighborEntityKeyId ),
-                            authorizedPropertyTypesOnNeighborES
+                            authorizedPropertyTypes.get( neighborEntitySetId )
                     );
                 }
             } );
         }
 
-        Map<UUID, PropertyType> authorizedPropertyTypesOnEntityES = authzHelper
-                .getAuthorizedPropertyTypes( entitySetId, WRITE_PERMISSION );
+        /*
+         * 2 - clear the given entity itself
+         */
+
         dgm.clearEntities(
                 entitySetId,
                 ImmutableSet.of( entityKeyId ),
-                authorizedPropertyTypesOnEntityES
+                authorizedPropertyTypes.get( entitySetId )
         );
 
         return null;
--- a/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
+++ b/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
@@ -57,6 +57,7 @@ import com.openlattice.datastore.services.EdmService;
 import com.openlattice.datastore.services.SearchService;
 import com.openlattice.datastore.services.SyncTicketService;
 import com.openlattice.edm.type.PropertyType;
+import com.openlattice.graph.Graph;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.olingo.commons.api.edm.EdmPrimitiveTypeKind;
 import org.apache.olingo.commons.api.edm.FullQualifiedName;
@@ -108,7 +109,7 @@ public class DataController implements DataApi, AuthorizingComponent {
     private SyncTicketService sts;
 
     @Inject
-    private EdmService dms;
+    private EdmService edmService;
 
     @Inject
     private DataGraphManager dgm;
@@ -127,6 +128,9 @@ public class DataController implements DataApi, AuthorizingComponent {
 
     @Inject
     private SearchService searchService;
+
+    @Inject
+    private Graph graph;
 
     private LoadingCache<UUID, EdmPrimitiveTypeKind>  primitiveTypeKinds;
     private LoadingCache<AuthorizationKey, Set<UUID>> authorizedPropertyCache;
@@ -210,7 +214,7 @@ public class DataController implements DataApi, AuthorizingComponent {
         final Map<UUID, PropertyType> authorizedPropertyTypes = authzHelper
                 .getAuthorizedPropertyTypes( entitySetId,
                         EnumSet.of( Permission.READ ),
-                        dms.getPropertyTypesAsMap( selectedProperties ) );
+                        edmService.getPropertyTypesAsMap( selectedProperties ) );
 
         final LinkedHashSet<String> orderedPropertyNames = new LinkedHashSet<>( authorizedPropertyTypes.size() );
 
@@ -248,9 +252,11 @@ public class DataController implements DataApi, AuthorizingComponent {
         if ( partialReplace ) {
             return dgm.partialReplaceEntities( entitySetId,
                     entities,
-                    dms.getPropertyTypesAsMap( requiredPropertyTypes ) );
+                    edmService.getPropertyTypesAsMap( requiredPropertyTypes ) );
         } else {
-            return dgm.replaceEntities( entitySetId, entities, dms.getPropertyTypesAsMap( requiredPropertyTypes ) );
+            return dgm.replaceEntities( entitySetId,
+                    entities,
+                    edmService.getPropertyTypesAsMap( requiredPropertyTypes ) );
         }
     }
 
@@ -268,7 +274,7 @@ public class DataController implements DataApi, AuthorizingComponent {
 
         return dgm.replacePropertiesInEntities( entitySetId,
                 entities,
-                dms.getPropertyTypesAsMap( requiredPropertyTypes ) );
+                edmService.getPropertyTypesAsMap( requiredPropertyTypes ) );
     }
 
     @Override
@@ -339,7 +345,7 @@ public class DataController implements DataApi, AuthorizingComponent {
         final SetMultimap<UUID, UUID> requiredPropertyTypes = requiredAssociationPropertyTypes( associations );
         accessCheck( aclKeysForAccessCheck( requiredPropertyTypes, WRITE_PERMISSION ) );
 
-        final Map<UUID, PropertyType> authorizedPropertyTypes = dms
+        final Map<UUID, PropertyType> authorizedPropertyTypes = edmService
                 .getPropertyTypesAsMap( ImmutableSet.copyOf( requiredPropertyTypes.values() ) );
         return associations.entrySet().stream().mapToInt( association -> {
             final UUID entitySetId = association.getKey();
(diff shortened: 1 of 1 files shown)


src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +417,51 @@ public Void clearEntitySet( UUID entitySetId ) {
         return null;
     }
 
+    @Override
+    @RequestMapping(
+            path = { "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + ENTITY_KEY_ID_PATH + "/graph" },
```
katherinebernstein: can we make this a constant?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +417,51 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+            method = RequestMethod.DELETE
+    )
+    public Void clearEntityAndNeighborEntities(
+            @PathVariable( ENTITY_SET_ID ) UUID entitySetId,
+            @PathVariable( ENTITY_ID ) UUID entityId
```
katherinebernstein: should be `entityKeyId` to avoid confusion with entity ids (the ones for integrations)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +417,51 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+    ) {
+
+        ensureReadAccess( new AclKey( entitySetId ) );
+        Map<UUID, List<NeighborEntityDetails>> result = searchService
+                .executeEntityNeighborSearch( entitySetId, ImmutableSet.of( entityId ) );
+
+        if ( result != null && !result.containsKey( entityId ) ) {
```
katherinebernstein: This should only continue if the result does contain the key, right?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +417,51 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                            .getAuthorizedPropertyTypes( neighborEntitySetId, WRITE_PERMISSION );
+                    dgm.clearEntities(
+                            neighborEntitySetId,
+                            ImmutableSet.of( neighborEntityId ),
+                            authorizedPropertyTypesOnNeighborES
+                    );
+                }
+            } );
+        }
+        return null;
```
katherinebernstein: Function name makes it sound like the entity itself will also be deleted, but it doesn't look like that's happening.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +417,51 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                .executeEntityNeighborSearch( entitySetId, ImmutableSet.of( entityId ) );
+
+        if ( result != null && !result.containsKey( entityId ) ) {
+
+            List<NeighborEntityDetails> neighbors = result.get( entityId );
+            neighbors.parallelStream().forEach( neighbor -> {
+
+                UUID associationEntityId = (UUID) neighbor.getAssociationDetails().get( ID_FQN ).iterator().next();
+                UUID associationEntitySetId = neighbor.getAssociationEntitySet().getId();
+                Map<UUID, PropertyType> authorizedPropertyTypesOnNeighborAES = authzHelper
```
katherinebernstein: Theoretically you can have write permissions on entity set properties without having write on the entity set itself -- unclear what we should do in this case. Do we need to also check for write permissions on the entity set? (Also not necessarily related to this PR, but what is expected behavior if someone calls delete on an entity and only has write permissions on some of the properties?)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +417,51 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                dgm.clearEntities(
+                        neighbor.getAssociationEntitySet().getId(),
+                        ImmutableSet.of( associationEntityId ),
+                        authorizedPropertyTypesOnNeighborAES
+                );
+
+                if ( neighbor.getNeighborId().isPresent() && neighbor.getNeighborEntitySet().isPresent() ) {
+                    UUID neighborEntityId = neighbor.getNeighborId().get();
+                    UUID neighborEntitySetId = neighbor.getNeighborEntitySet().get().getId();
+                    Map<UUID, PropertyType> authorizedPropertyTypesOnNeighborES = authzHelper
```
katherinebernstein: A lot of these auth checks are probably going to be repeated a bunch of times. It makes the code less pretty, but it would probably speed this up a lot to do an initial pass to check permissions, avoid repeating access checks, and then leverage the bulk delete. (I know for the PSA there are a bunch of people with hundreds of associated charges, so this would save thousands of unnecessary calls each time)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
katherinebernstein: (same for `neighborEntityId` and `associationEntityId`)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +417,51 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                Map<UUID, PropertyType> authorizedPropertyTypesOnNeighborAES = authzHelper
+                        .getAuthorizedPropertyTypes( associationEntitySetId, WRITE_PERMISSION );
+                dgm.clearEntities(
+                        neighbor.getAssociationEntitySet().getId(),
+                        ImmutableSet.of( associationEntityId ),
+                        authorizedPropertyTypesOnNeighborAES
+                );
+
+                if ( neighbor.getNeighborId().isPresent() && neighbor.getNeighborEntitySet().isPresent() ) {
+                    UUID neighborEntityId = neighbor.getNeighborId().get();
```
katherinebernstein: We might want to switch to using `ID_FQN` here too -- probably best to move away from using `getNeighborId` in general although once we rewrite neighbors we'll probably need to rewrite this anyway, so it might not matter.

katherinebernstein (changes requested):

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Ooops. It's a constant in `api`, just forgot to update here. Thanks!

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: What's the difference between `entityId` and `entityKeyId`? I thought `entityId` referred to the id of the actual piece of data, the entity?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Good catch.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Fixed.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: I thought about that too. My plan was to get the basic logic down in the first implementation (this), and refactor after a few reviews.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Not sure what the behavior should be; definitely worth discussing in longer form.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Ok, refactored the code to avoid repeating calls to `getAuthorizedPropertyTypes()`.

geekbeast (approved):

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,101 @@ public Void clearEntitySet( UUID entitySetId ) {
         return null;
     }
 
+    @Override
+    @RequestMapping(
+            path = { "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + ENTITY_KEY_ID_PATH + "/" + NEIGHBORS },
+            method = RequestMethod.DELETE
+    )
+    public Void clearEntityAndNeighborEntities(
```
geekbeast: Even if you are going to expose the singleton endpoint you should also expose a bulk delete end point and have the singleton endpoint invoke it.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,101 @@ public Void clearEntitySet( UUID entitySetId ) {
         return null;
     }
 
+    @Override
+    @RequestMapping(
+            path = { "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + ENTITY_KEY_ID_PATH + "/" + NEIGHBORS },
```
geekbeast: Not a fan of this being in the DataController as the paths are starting to get dirty from all the operations we are putting on it. Understand if this is necessary for now.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,101 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+            method = RequestMethod.DELETE
+    )
+    public Void clearEntityAndNeighborEntities(
+            @PathVariable( ENTITY_SET_ID ) UUID entitySetId,
+            @PathVariable( ENTITY_KEY_ID ) UUID entityKeyId
+    ) {
+
+        ensureReadAccess( new AclKey( entitySetId ) );
```
geekbeast: You should perform all access checks in a single batch.  I think some of the create association calls should have examples patterns you can take a look at.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,101 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+        authorizedPropertyTypes.put(
+                entitySetId,
+                authzHelper.getAuthorizedPropertyTypes( entitySetId, WRITE_PERMISSION )
+        );
+
+        /*
+         * 1 - get the entity neighbors
+         */
+
+        Map<UUID, List<NeighborEntityDetails>> result = searchService
```
geekbeast: This call is crazy big and unnecessary. It's more appropriate to add an API on the DataGraphManager that calls the Graph class.

geekbeast: i.e we want to delete this API down the road, because the API itself is flawed.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,101 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                .executeEntityNeighborSearch( entitySetId, ImmutableSet.of( entityKeyId ) );
```
geekbeast: Invoked from datastore this triggers a ton of binary serialization (another reason to not use the API).

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,101 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                }
+            } );
+
+            /*
+             * 1.2 - clear all neighbor and association entities
+             */
+
+            neighborEntitiesMap.keySet().forEach( neighborEntitySetId -> {
+                Set<UUID> neighborEntityKeyIds = neighborEntitiesMap.get( neighborEntitySetId );
+                dgm.clearEntities(
```
geekbeast: Another major thing here is that graph edges are versioned so implementing this properly requires clearing versions in the edges as well as this is different from a a delete (clear only tombstones, while delete also deletes edges).

geekbeast (changes requested):

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: The client will need to trigger a delete on the entity and its neighbors. How do you recommend exposing that if not via the API?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Fixed. (I think)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Fixed. (I think)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Fixed. (I think)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
geekbeast: This is a comment about paths/controllers not the mechanism.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +424,87 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+            @PathVariable( ENTITY_KEY_ID ) UUID entityKeyId
+    ) {
+
+        /*
+         * 1 - check permissions for the given EntitySet as well as all of its neighbor EntitySets
+         */
+
+        // getNeighborEntitySetIds() returns source, destination, and edge EntitySet ids
+        Set<UUID> neighborEntitySetIds = dgm.getNeighborEntitySetIds( entitySetId );
+        Set<UUID> allEntitySetIds = ImmutableSet.<UUID>builder()
```
geekbeast: Just use neighborEntitySetIds.stream() and Stream.concat to avoid re-hashing every object in the entity set.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +424,87 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                .add( entitySetId )
+                .addAll( neighborEntitySetIds )
+                .build();
+
+        Map<AclKey, EnumSet<Permission>> entitySetsAccessRequestMap = allEntitySetIds
+                .parallelStream()
```
geekbeast: Parallel streams have high overhead it's probably not worth it for just AclKey::new and then you don't then you can Collectors.toMap(...)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +424,87 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+
+        entitySetIdToEntityKeyIdsMap.keySet().parallelStream().forEach( esId -> {
+                dgm.clearEntities(
+                        esId,
+                        entitySetIdToEntityKeyIdsMap.get( esId ),
+                        entitySetIdToPropertyTypesMap.get( esId )
+                );
+        } );
+
+        return null;
```
geekbeast: Return number of items cleared/deleted.

geekbeast (changes requested):

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: These are UUIDs, not whole EntitySet objects. Is there really a major difference between `Stream.concat` and `.add` / `.addAll`?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Fixed.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: Integer or Long?

UnsungHero97: Fixed.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,82 @@ public Void clearEntitySet( UUID entitySetId ) {
         return null;
     }
 
+    @Override
+    @Timed
+    @RequestMapping(
+            path = { "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + ENTITY_KEY_ID_PATH + "/" + NEIGHBORS },
+            method = RequestMethod.DELETE
+    )
```
geekbeast, line 426: Is this from auto-formatter?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,82 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+            @PathVariable( ENTITY_SET_ID ) UUID vertexEntitySetId,
+            @PathVariable( ENTITY_KEY_ID ) UUID vertexEntityKeyId
+    ) {
+
+        /*
+         * 1 - collect all relevant EntitySets and PropertyTypes
+         */
+
+        // getNeighborEntitySetIds() returns source, destination, and edge EntitySet ids
+        Set<UUID> neighborEntitySetIds = dgm.getNeighborEntitySetIds( vertexEntitySetId );
```
geekbeast, line 437: Is this full copy necessary?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -414,6 +418,82 @@ public Void clearEntitySet( UUID entitySetId ) {
...
+                    entitySetIdToPropertyTypesMap.put( esId, propertyTypes );
+                    return entitySetIdToPropertyTypesMap.keySet().stream().map( ptId -> new AclKey( esId, ptId ) );
+                } )
+                .collect( Collectors.toMap( Function.identity(), aclKey -> WRITE_PERMISSION ) );
+
+        /*
+         * 2 - check permissions for all relevant EntitySets and PropertyTypes
+         */
+
+        Map<AclKey, EnumSet<Permission>> accessRequestMap = Maps.newHashMap();
```
geekbeast, line 462: Unnecessary double copy.

geekbeast (approved):
There's some minor changes that are nice to have.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: What do you mean full copy? Are you suggesting passing the result from `getNeighborEntitySetIds()` directly to the `.addAll()`?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
UnsungHero97: What do you mean double copy? Are you suggesting not creating `accessRequestMap`, and instead somehow passing the 2 maps directly into `accessCheck()`?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
geekbeast: allEntitySetIds = neighborEntitySetIds + vertexEntitySetId but is constructed by rehashing all neighborEntitySetIds, which returns a mutable set safe to use set. If you are concerned about safe guarding accidental mutation of the set ```Collections.unmodifiableSet(...)``` seems more appropriate in this case.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
geekbeast: accessRequestMap = entitySetsAccessRequestMap + propertyTypesAccessRequestMap

You can avoid a lot of the work here by ```Stream<AclKey> entitySetsAccessRequests = ...``` and ```Stream<AclKey> propertyTypeAccessRequests = ...``` and then ```Stream.concat(entitySetsAccessRequests, propertyTypeAccessRequest).collect( Function.identity(), aclKey -> aclKey.size() == 1 ? READ_PERMISSION : WRITE_PERMISSION)```

geekbeast:
Merging for greater good.

PR #283, opened 2018-09-14
changed implementation of getAclExplanation

pckeyes:
Implemented a search of the principals tree to create permissions paths. I think it is only part of what needs to be done for getAclExplanation, but I'm looking for feedback as to whether I'm on the right track.


src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -125,50 +117,117 @@ public Acl getAcl( @RequestBody AclKey aclKeys ) {
...
+
+            principalToPrincipalPaths.put( principal, paths );
+
+            //get parent layer
+            Collection<SecurablePrincipal> parentLayer = securePrincipalsManager
+                    .getParentPrincipalsOfPrincipal( securePrincipalsManager.lookup( principal ) );
+            //if parent layer isn't empty, start recursively getting paths
+            if ( !parentLayer.isEmpty() ) {
+                for ( SecurablePrincipal parent : parentLayer ) {
+                    paths = securePrincipalsManager
```
geekbeast: Why is this in SPM?

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -125,50 +117,117 @@ public Acl getAcl( @RequestBody AclKey aclKeys ) {
...
+            //OffsetDateTime expirationDate = ace.getExpirationDate();
+            List<List<Principal>> paths = new ArrayList<>();
+
+            principalToPrincipalPaths.put( principal, paths );
+
+            //get parent layer
+            Collection<SecurablePrincipal> parentLayer = securePrincipalsManager
+                    .getParentPrincipalsOfPrincipal( securePrincipalsManager.lookup( principal ) );
+            //if parent layer isn't empty, start recursively getting paths
+            if ( !parentLayer.isEmpty() ) {
```
geekbeast: Not necessary.

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -125,50 +117,117 @@ public Acl getAcl( @RequestBody AclKey aclKeys ) {
...
+                .create(); //will hold all aces that contain the key principal
+
+        Iterable<Ace> aces = authorizations.getAllSecurableObjectPermissions( aclKey )
+                .getAces(); //gets aces from returned acl
+        Map<PrincipalType, List<Ace>> acesByType = StreamUtil.stream( aces )
+                .collect( Collectors.groupingBy( ace -> ace.getPrincipal().getType() ) );
+        //for principals that aren't users, sPM.getallPrincipalsbyPrincipal gets all of the parents of those principals
+        Map<Principal, List<List<Principal>>> principalToPrincipalPaths = new HashMap<>();
+        //for all principals with direct access
+        for ( Ace ace : aces ) {
```
geekbeast: Not sure aces can be re-used... some of our Iterable are one time use only for legacy reasons.

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -125,50 +117,117 @@ public Acl getAcl( @RequestBody AclKey aclKeys ) {
...
+        ensureOwnerAccess( aclKey );
+        SetMultimap<Principal, Ace> resultMap = HashMultimap
+                .create(); //will hold all aces that contain the key principal
+
+        Iterable<Ace> aces = authorizations.getAllSecurableObjectPermissions( aclKey )
+                .getAces(); //gets aces from returned acl
+        Map<PrincipalType, List<Ace>> acesByType = StreamUtil.stream( aces )
+                .collect( Collectors.groupingBy( ace -> ace.getPrincipal().getType() ) );
+        //for principals that aren't users, sPM.getallPrincipalsbyPrincipal gets all of the parents of those principals
+        Map<Principal, List<List<Principal>>> principalToPrincipalPaths = new HashMap<>();
```
geekbeast: You should initialize this from acesByType. acesByType.entrySet().stream().filter( e->e.getKey()!=PrincipalType.USER ).flatMap( e-> e.getValue().stream() ).collect( ace -> ace.getPrincipal(), Lists.newArrayList( Lists.newArrayList() ))_

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
geekbeast: for( Entry<Principal, List<List<Principal>> e : principalToPrincipalPaths ) {
 var parentLayer = securePrincipalsManager.getParentPrincipalsOfPrincipal( ... );
 for( SecurablePrincipal parent : parentLayer ) {
    principalsToPrincipalPaths.merge( parent.getPrincipal(), e.getValue().stream().map( pl -> pl.add(e.getKey() ) ).collect(Collectors.toList()), (p1,p2)->{p1.addAll(p2); return p1;} );

you might also have to move switch Set<List<Principal>> since duplicate paths might get added. It's not great, but haven't had time to look at it closer.

geekbeast (changes requested):

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
pckeyes: The recursive function? It seemed to be performing a similar task as the other methods there. I can move it though, where should it go?

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
pckeyes: The if statement or this entire block of code?

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
pckeyes: Won't the result of calling parent.getPrincipal() return keys that aren't in the map yet?

PR #294, opened 2018-10-12
Feature/lattice 1473 read linked data

you-leee:
First bigger change of reading linking entity sets.

### DIFF
--- a/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
+++ b/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
@@ -48,9 +48,8 @@ import com.openlattice.datastore.constants.CustomMediaType;
 import com.openlattice.datastore.services.EdmService;
 import com.openlattice.datastore.services.SyncTicketService;
 import com.openlattice.edm.EntitySet;
-import com.openlattice.edm.type.EntityType;
 import com.openlattice.edm.type.PropertyType;
-import com.openlattice.linking.LinkingQueryService;
+import com.openlattice.linking.LinkingEntitySetQueryService;
 import com.openlattice.search.SearchService;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.olingo.commons.api.edm.EdmPrimitiveTypeKind;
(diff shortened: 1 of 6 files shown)


src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -189,12 +189,55 @@
...
+    private EntitySetData<FullQualifiedName> loadLinkedEntitySetData(
+            UUID linkingEntitySetId,
+            EntitySetSelection selection,
+            EntitySet es) {
+        Set<UUID> allEntitySetIds = Sets.newHashSet( es.getLinkedEntitySets());
+        checkState( !allEntitySetIds.isEmpty(),
+                "Linked entity sets are empty for linking entity set %s", linkingEntitySetId );
+
+        EntityType linkedEntityType = edmService.getEntityTypeByEntitySetId( allEntitySetIds.iterator().next() );
+        checkState( allEntitySetIds.stream().skip( 1 )
```
geekbeast: We shouldn't check this for every read. We should ensure in the linking controller that you can't add an entity set of the wrong type. This adds overhead to every single read call.

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -189,12 +189,55 @@
...
+            Set<UUID> allEntitySetIds,
+            Optional<Set<UUID>> linkingIds) {
+
+        if(linkingIds.isPresent()) {
+            // if linkingIds is there: get all entityKey ids per entity set where linking_id =
+            Map<UUID, Set<UUID>> entityKeyIdsOfLinkingIds = linkingQueryService.getEntitiesWithLinkingIds(
+                    linkingIds.get());
+
+            return allEntitySetIds.stream().collect(Collectors.toMap(
+                    esId -> esId,
```
geekbeast: Function.identity()

geekbeast: Actually why not just ImmutableMap.copyOf(Maps.transformValues( entityKeyIdsOfLinkingIds, Optional::of) )?

geekbeast (changes requested):

PR #307, opened 2018-11-16
endpoint to clear all entity set data without deleting the entity set

katherinebernstein:


src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -440,6 +441,23 @@ public Void clearEntityFromEntitySet(
         return null;
     }
 
+    @Override
+    @RequestMapping(
+            path = { "/" + ENTITY_SET + "/" + SET_ID_PATH + "/" + ENTITIES },
+            method = RequestMethod.DELETE )
+    public Integer clearAllEntitiesFromEntitySet( @PathVariable( ENTITY_SET_ID ) UUID entitySetId ) {
```
geekbeast, line 448: You should make sure that it is not a linking entity set as linking entity sets can't be be erased.

geekbeast (changes requested):

geekbeast (approved):

PR #322, opened 2018-12-19
access check id fixes

katherinebernstein:


geekbeast (commented):
I looked at getAuthorizedPropertyTypes and getAuthorizedPropertiesOnEntitySet and the root cause I think in EdmAuthorizationHelper line 71 has a bug. 

It should be ```.filter( authz -> authz.getValue().values().stream().anyMatch( v -> !v ) )``` instead of ```.filter( authz -> authz.getValue().values().stream().allMatch( v -> !v ) )```.

The latter will remove the property type only if all permission checks came back false (instead of if any of the requested permissions weren't granted). I don't know if this is inline with the bug Kim submitted, but we seem to have multiple ways of getting authorized property types.

The bigger issue is that we seem to have way too many ways of getting the authorized set of property types and they do different things.

PR #326, opened 2019-01-06
Bugfix/lattice 1721 self aggregations

geekbeast:


geekbeast:
Snuck in a fix for LATTICE-58

PR #331, opened 2019-01-24
LATTICE-1788 add check for environment before clearing tables, fix co…

you-leee:
…mpile

### DIFF
--- a/src/main/java/com/openlattice/datastore/edm/controllers/EdmController.java
+++ b/src/main/java/com/openlattice/datastore/edm/controllers/EdmController.java
@@ -20,6 +20,8 @@
 
 package com.openlattice.datastore.edm.controllers;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/openlattice/datastore/edm/controllers/EdmController.java
```diff
@@ -112,11 +115,18 @@
...
 
+    @Inject
+    private Environment env;
+
     @RequestMapping(
             path = CLEAR_PATH,
             method = RequestMethod.DELETE )
     @ResponseStatus( HttpStatus.OK )
     public void clearAllData() {
+        if ( !env.acceptsProfiles( TEST_PROFILE ) ) {
```
geekbeast, line 126: I still think we should move this to a different controller and api, but good enough for now.

geekbeast (approved):

PR #350, opened 2019-02-26
LATTICE-1805 move role owner check 1 level up when adding role to user

you-leee:


src/main/java/com/openlattice/controllers/OrganizationsController.java
```diff
@@ -445,9 +445,9 @@ public Void addRoleToUser(
             @PathVariable( ROLE_ID ) UUID roleId,
             @PathVariable( USER_ID ) String userId ) {
         ensureWriteAccess( new AclKey( organizationId ) );
```
katherinebernstein: I think the decision was that you don't need write access on the organization to add/remove a role to/from a user -- only need owner access on the role

you-leee: I think we agreed on not needing owner access on organization, only read (and owner access on role). Read check on organization was there also before.

katherinebernstein: I think you shouldn't need any permission on the organizations. Since there's no inheritance through AclKeys, for managing permissions on any AclKey of the form `[id1, id2]`, all you need is owner permissions on the AclKey `[id1, id2]` and no required permissions on the AclKey `[id1]` (whether that's roles in an organization, or property types on an entity set)

you-leee: Okay, I will remove the check. Thanks for the explanation!!

geekbeast (approved):

PR #359, opened 2019-03-13
LATTICE-1883 check for materialize on entity set, not on properties, …

you-leee:
…when materializing entity sets

### DIFF
--- a/src/main/java/com/openlattice/controllers/OrganizationsController.java
+++ b/src/main/java/com/openlattice/controllers/OrganizationsController.java
@@ -40,6 +40,7 @@ import com.openlattice.controllers.exceptions.ForbiddenException;
 import com.openlattice.controllers.exceptions.ResourceNotFoundException;
 import com.openlattice.datastore.services.EdmManager;
 import com.openlattice.directory.pojo.Auth0UserBasic;
+import com.openlattice.edm.type.PropertyType;
 import com.openlattice.organization.Organization;
 import com.openlattice.organization.OrganizationIntegrationAccount;
 import com.openlattice.organization.OrganizationMember;
@@ -53,6 +54,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


src/main/java/com/openlattice/controllers/OrganizationsController.java
```diff
@@ -212,14 +208,22 @@ public OrganizationIntegrationAccount getOrganizationIntegrationAccount( @PathVa
...
-                authzHelper.getAuthorizedPropertiesOnEntitySets( entitySetIds, EnumSet.of( Permission.MATERIALIZE ) );
+
         final var organizationPrincipal = organizations.getOrganizationPrincipal( organizationId );
         if ( organizationPrincipal == null ) {
             //This will be rare, since it is unlikely you have access to an organization that does not exist.
             throw new ResourceNotFoundException( "Organization does not exist." );
         }
+
+        entitySetIds.forEach( entitySetId -> ensureMaterialize(entitySetId, organizationPrincipal) );
+        final var authorizedPropertyTypesByEntitySet = authzHelper.getAuthorizedPropertiesOnEntitySets(
```
katherinebernstein, line 222: I don't think we want to check materialize on the the property types -- as long as the person has materialize permissions on the entity set aclKey, the full entity set should be materialized, and then the correct permissions will be enforced on reading the materialized entity set. Since the organization shares a single materialized view for an entity set, partially materializing will lead to problems for people who do have full permissions.

katherinebernstein (changes requested):

src/main/java/com/openlattice/controllers/OrganizationsController.java
you-leee: I followed the documentation on this, but maybe I misunderstood: https://wiki.openlattice.com/pages/viewpage.action?spaceKey=ENGR&title=Materialized+Entity+Set+Views
As I understand: the owner of org can add/revoke materialize permissions to/from the org principal

katherinebernstein: oh you're right -- this may currently be problematic since I don't think there's any way to grant permissions to an organization right now (other than postman) so nothing will be materializable until we make those updates. Also we are currently granting / making people able to grant materialize to users and roles, and removing that may introduce some new problems in cases where we're checking against EnumSet.allOf( Permission.class) , etc

katherinebernstein (approved):

PR #373, opened 2019-04-04
LATTICE-1915 Deleting all entities from an entity set enumerates all entityKeyIds

makosblade:
This PR:
- Updates `deleteEntities` to call a more efficient method when deleting an entire entity set

Notes:
- Unsure whether this is the correct fix since the audit logging will say that `deleteAllEntitiesFromEntitySet` was used rather than `deleteEntities`
- May be better to just return a 400 and recommend calling `deleteAllEntitiesFromEntitySet`

### DIFF
--- a/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
+++ b/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
@@ -670,6 +670,11 @@ public class DataController implements DataApi, AuthorizingComponent, AuditingCo
             @PathVariable( ENTITY_SET_ID ) UUID entitySetId,
             @RequestBody Set<UUID> entityKeyIds,
             @RequestParam( value = TYPE ) DeleteType deleteType ) {
+
+        if ( entityKeyIds.size() == dgm.getEntitySetSize( entitySetId ) ) {
+            return deleteAllEntitiesFromEntitySet( entitySetId, deleteType );
+        }
+
         WriteEvent writeEvent;
         if ( deleteType == DeleteType.Hard ) {
             // access checks for entity set and properties
--- a/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
+++ b/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
@@ -671,8 +671,8 @@ public class DataController implements DataApi, AuthorizingComponent, AuditingCo
             @RequestBody Set<UUID> entityKeyIds,
             @RequestParam( value = TYPE ) DeleteType deleteType ) {
 
-        if ( entityKeyIds.size() == dgm.getEntitySetSize( entitySetId ) ) {
-            return deleteAllEntitiesFromEntitySet( entitySetId, deleteType );
+        if ( entityKeyIds.size() > MAX_BATCH_SIZE ) {
+            throw new IllegalArgumentException( "You can only delete entities in batches of up to " + MAX_BATCH_SIZE + " per request." );
         }
 
         WriteEvent writeEvent;


src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -670,6 +670,11 @@ public Integer deleteEntities(
             @PathVariable( ENTITY_SET_ID ) UUID entitySetId,
             @RequestBody Set<UUID> entityKeyIds,
             @RequestParam( value = TYPE ) DeleteType deleteType ) {
+
+        if ( entityKeyIds.size() == dgm.getEntitySetSize( entitySetId ) ) {
+            return deleteAllEntitiesFromEntitySet( entitySetId, deleteType );
```
katherinebernstein: This will probably work in pretty much every case, but it's theoretically possible that `entityKeyIds` could include an entityKeyId that was already deleted, meaning it wouldn't be included in the entity set size computed from DataGraphManager, which would result in extra entities being unintentionally deleted

makosblade: Oof, yeah you're totally right.
In that case though, I don't see how we could fix this aside from enforcing a max size on `entityKeyIds` which could still result in being able to remove an entire entity set through here but wouldn't be a big deal.

makosblade: I believe the default is 128MB which would be like 8 million UUIDs

you-leee: I would go with enforcing too or doing it in chunks

you-leee (approved):

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -670,6 +670,11 @@ public Integer deleteEntities(
             @PathVariable( ENTITY_SET_ID ) UUID entitySetId,
             @RequestBody Set<UUID> entityKeyIds,
             @RequestParam( value = TYPE ) DeleteType deleteType ) {
+
```
katherinebernstein, line 673: unrelated but just noticed we're not calling `ensureWriteAccess` here which we should definitely be doing... (also that we call `ensureReadAccess` if it's a soft delete, which we should not be doing)

makosblade: I thought that `ensureReadAccess` looked shifty when I first read this

makosblade: LATTICE-1924

katherinebernstein: cool thanks!

you-leee: This is a pretty old debate actually, but last time I asked Matthew for permission checks on soft delete, the answer was this:
"Julia [15:55]
in soft deletes, which kind of permissions are required? on entityset and on properties?
because here I’m seeing, that for entityset it’s only checking read
and properties only write

matthew [16:12]
That's correct
soft delete is a lot more permissive
because it doesn't remove data"

katherinebernstein: yeah +1 since we're checking property type permissions and permission inheritance isn't a thing -- although why is read needed on the entity set?

you-leee: I'm not sure.. maybe because this is basically a write to the entity set?

geekbeast: You need read on entity set metadata to issue any mutations against an entity set that rely upon that data (which is pretty much anything except maybe reading a specific property).

geekbeast (approved):

geekbeast:
Pattern ensureRead( entitySetId ) -> ensureWrite( properties ) -> do work

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
you-leee: So, if the pattern is first checking permissions, it shouldn't be allowed to clean only authorized properties? Right now it's working like that

PR #382, opened 2019-04-27
Feature/lattice 1929 unify permission handling linking entitysets

you-leee:


geekbeast (approved):

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -141,8 +143,53 @@ public SearchResult getEntitySets(
             produces = { MediaType.APPLICATION_JSON_VALUE } )
     @Override
     public DataSearchResult searchEntitySetData( @RequestBody SearchConstraints searchConstraints ) {
-
-        DataSearchResult results = searchService.executeSearch( searchConstraints, Principals.getCurrentPrincipals() );
+        // check read on entity sets
+        final var authorizedEntitySetIds = Arrays.stream( searchConstraints.getEntitySetIds() )
```
katherinebernstein: Is all this authorization logic also present in SearchService? `SearchService. executeSearch` is also called from `PersistentSearchMessenger` for a particular user's principals, so authorization piece was originally moved into SearchService so that we didn't have to duplicate all this authorization logic in multiple places.

you-leee: I moved it out of SearchService.executeSearch to the SearchController and PersistentSearchMessengerTask, because far as I know, we prefer to do all the authentication in the controller classes and the logic for linking entity sets had to change anyway. 
Now SearchService.executeSearch takes only SearchConstraints searchConstraints and Map<UUID, Map<UUID, PropertyType>> authorizedPropertyTypesByEntitySet as inputs

katherinebernstein: In that case I think it would be better to either move the logic back into SearchService or (preferable) add something like getAuthorizedPropertyTypesByEntitySetsForPrincipals that performs this logic in HazelcastAuthorizationService to avoid having to maintain this logic in multiple places.

you-leee: Okay, I will move it back then

katherinebernstein: I think moving it into HazelcastAuthorizationService would be better actually -- this seems like something that might be useful in other places than this function

you-leee: Okay, I will create a function in EdmAuthorizationHelper

katherinebernstein: Oh yeah that would be better than HazelcastAuthorizationService. Cool thanks!

you-leee: One question though: In SearchController.searchEntitySetData shouldn't we only allow the search if the calling user has read access on all of them? Or if we allow partial searches, shouldn't we change the code in SearchService.executeSearch to only do the search on authorized entitysets (right now it's using the entity set ids of the SearchConstraints)

katherinebernstein: Yeah, I think it makes sense require read on all the entity sets but not require read on all property types

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -240,13 +273,12 @@ public DataSearchResult executeAdvancedEntitySetDataQuery(
...
         DataSearchResult dataSearchResult = new DataSearchResult( 0, Lists.newArrayList() );
 
-        Set<Principal> currentPrincipals = Principals.getCurrentPrincipals();
-
         if ( authorizations.checkIfHasPermissions( new AclKey( entitySetId ),
                 Principals.getCurrentPrincipals(),
                 EnumSet.of( Permission.READ ) ) ) {
 
             EntitySet es = edm.getEntitySet( entitySetId );
+
```
katherinebernstein: you can actually just get rid of everything in this function and just call through to `searchEntitySetData` with the `SearchConstraints.advancedSearchConstraints` object below (and the difference in auditing doesn't matter since the SearchType will be recorded)

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -192,27 +219,33 @@ public DataSearchResult executeEntitySetDataQuery(
         if ( authorizations.checkIfHasPermissions( new AclKey( entitySetId ),
                 currentPrincipals,
                 EnumSet.of( Permission.READ ) ) ) {
+
```
katherinebernstein: same comment as for the advanced search

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -141,8 +141,36 @@ public SearchResult getEntitySets(
             produces = { MediaType.APPLICATION_JSON_VALUE } )
     @Override
     public DataSearchResult searchEntitySetData( @RequestBody SearchConstraints searchConstraints ) {
-
-        DataSearchResult results = searchService.executeSearch( searchConstraints, Principals.getCurrentPrincipals() );
+        // check read on entity sets
+        final var allNormalEntitySetIds = Arrays.stream( searchConstraints.getEntitySetIds() )
```
katherinebernstein: This logic doesn't look present in the PersistentSearchMessenger change in conductor-client. The permission flow here and there should be identical given a set of principals, so this might be something to add to EdmAuthorizationHelper too

you-leee: Thanks, my bad, will include entity set read checks there also.
I will add a different call, where it checks some permissions on entitysets, but I wouldn't include it in the same call, because that is supposed to check for propertytypes.

katherinebernstein: sounds good!

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -1025,31 +1011,46 @@ public long getEntitySetSize( @PathVariable( ENTITY_SET_ID ) UUID entitySetId )
...
+
+            final Map<UUID, Map<UUID, PropertyType>> authorizedPropertyTypes = authzHelper
+                    .getAuthorizedPropertyTypesByNormalEntitySet(
+                            entitySet,
                             Set.of( propertyTypeId ),
                             EnumSet.of( Permission.READ ) );
-            FullQualifiedName propertyTypeFqn = authorizedPropertyTypes.get( entitySetId ).get( propertyTypeId )
+
+            // if any of its normal entitysets don't have read permission on property type, reading is not allowed
+            if ( authorizedPropertyTypes.values().iterator().next().isEmpty() ) {
```
katherinebernstein, line 1033: shouldn't this be `authorizedPropertyTypes.values().iterator().next().containsKey( propertyTypeId )` ?

you-leee: The getAuthorizedPropertyTypesByNormalEntitySet takes the property type as argument, so, if the calling user has no access, it should be empty, but that would work too.

PR #384, opened 2019-05-01
Add a check in EdmController for EdmPrimitiveTypeKind.Binary Property…

makosblade:
…Types with IndexTypes other than IndexType.NONE

This PR:
- Adds an exceptional case to EdmController when a PropertyType creation is attempted on Binary data with an IndexType other than 'none'

### DIFF
--- a/src/main/java/com/openlattice/datastore/edm/controllers/EdmController.java
+++ b/src/main/java/com/openlattice/datastore/edm/controllers/EdmController.java
@@ -45,12 +45,17 @@ import com.openlattice.edm.requests.EdmRequest;
 import com.openlattice.edm.requests.MetadataUpdate;
 import com.openlattice.edm.schemas.manager.HazelcastSchemaManager;
 import com.openlattice.edm.set.EntitySetPropertyMetadata;
-import com.openlattice.edm.type.*;
+import com.openlattice.edm.type.AssociationDetails;
+import com.openlattice.edm.type.AssociationType;
+import com.openlattice.edm.type.EntityType;
+import com.openlattice.edm.type.PropertyType;
 import com.openlattice.organizations.roles.SecurePrincipalsManager;
+import com.openlattice.postgres.IndexType;
 import com.openlattice.postgres.streams.PostgresIterable;
 import com.openlattice.web.mediatypes.CustomMediaType;
 import org.apache.commons.lang3.NotImplementedException;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.olingo.commons.api.edm.EdmPrimitiveTypeKind;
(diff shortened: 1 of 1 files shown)


src/main/java/com/openlattice/datastore/edm/controllers/EdmController.java
```diff
@@ -787,6 +792,10 @@ public Void deleteEntityType( @PathVariable( ID ) UUID entityTypeId ) {
     @ResponseStatus( HttpStatus.OK )
     public UUID createPropertyType( @RequestBody PropertyType propertyType ) {
         ensureAdminAccess();
+
+        if ( EdmPrimitiveTypeKind.Binary == propertyType.getDatatype() && propertyType.getPostgresIndexType() != IndexType.NONE ){
+            throw new BadRequestException( "Indexes are not allowed on Binary property types" );
```
katherinebernstein, line 797: I think it's probably better to include this in the PropertyType constructor in case property types are ever created through some path other than this endpoint

makosblade: That's what I was thinking initially (and did that w/out the exception) but since the PropertyType constructor is a JSONCreator / used for deserialization, I think if we create a Binary property type and then explicitly update its indexType, then later when its deserialized elsewhere we'll get explosions.

katherinebernstein: True, although including it in the constructor would prevent it from being written into postgres (so deserialization should **theoretically** never happen) and would make it more visible if there's some service trying to create an index on binary data / expecting that index to exist. But, also true about exploding things if we get into a weird state

makosblade: Hmm. I may have misunderstood part of this ticket 😛 
Can you take a look at this: https://github.com/openlattice/conductor-client/pull/463 ?
My thought was that we would potentially want to intentionally set the index type on binary data later on to something other than `NONE`

katherinebernstein: Well all the binary data is actually stored in S3 and not in postgres, so there's not really anything to create an index on

makosblade: Ahh, okay. I'll just get rid of this portion then!

makosblade: Err, above referenced portion

makosblade:
Closing, constructing PropertyTypes ensures no Indexed binary data in this PR: https://github.com/openlattice/api/pull/320/files

PR #385, opened 2019-05-04
delete in batches

katherinebernstein:


src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -1143,6 +1062,44 @@ private static OffsetDateTime getDateTimeFromLong( long epochTime ) {
...
+            if ( deleteType == DeleteType.Hard ) {
+                deleteAssociations( entitySetId, Optional.of( chunk ) );
+                writeEvent = dgm.deleteEntities(
+                        entitySetId,
+                        chunk,
+                        authorizedPropertyTypes );
+            } else {
+                List<WriteEvent> clearEvents = clearAssociations( entitySetId, Optional.of( chunk ) );
+                while ( !clearEvents.isEmpty() ) {
+                    clearEvents = clearAssociations( entitySetId, Optional.of( chunk ) );
```
UnsungHero97: This looks identical to the one above. Why do you have to call `clearAssociations` in a loop with the exact same parameters after you've already called it once before?

katherinebernstein: Ooh thanks good catch, this was left over from a previous iteration

UnsungHero97 (approved):

geekbeast (approved):

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -1154,6 +1073,41 @@ private static OffsetDateTime getDateTimeFromLong( long epochTime ) {
...
+            } else {
+                clearAssociations( entitySetId, Optional.of( chunk ) );
+                writeEvent = dgm.clearEntities(
+                        entitySetId,
+                        chunk,
+                        authorizedPropertyTypes );
+            }
+
+            numUpdates += writeEvent.getNumUpdates();
+            maxVersion = Math.max( maxVersion, writeEvent.getVersion() );
```
geekbeast, line 1105: Version doesn't matter for hard deletes. For soft deletes using max version instead of min version will mean an auditor sees the state of the system after a bunch of operations instead of before.

The preferred way would seem to be to use one single version captured from before when the deletes started issuing. 

We should also consider that these operations are not a single transaction and should be audited as a compound operation or modified such that we issue all deletes in a single transaction.

PR #386, opened 2019-05-08
LATTICE-1990 add missing authorization check to DataIntegrationContro…

you-leee:
…ller, add EdmManager to DataGraphManager

### DIFF
--- a/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
+++ b/src/main/java/com/openlattice/datastore/data/controllers/DataController.java
@@ -309,7 +309,8 @@ public class DataController implements DataApi, AuthorizingComponent, AuditingCo
     @Timed
     @PutMapping( value = "/" + ASSOCIATION, consumes = MediaType.APPLICATION_JSON_VALUE )
     public Integer createAssociations( @RequestBody Set<DataEdgeKey> associations ) {
-        final var associationEntitySetIds = new HashMap<UUID, Map<UUID, Set<UUID>>>(); // edge-src-dst
+        final var srcAssociationEntitySetIds = new HashMap<UUID, Set<UUID>>(); // edge-src
+        final var dstAssociationEntitySetIds = new HashMap<UUID, Set<UUID>>(); // edge-dst
 
         associations.forEach(
                 association -> {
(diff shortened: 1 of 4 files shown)


src/main/java/com/openlattice/datastore/data/controllers/DataIntegrationController.java
```diff
@@ -216,7 +219,40 @@ public IntegrationResults integrateEntityAndAssociationData(
...
+                    ensureReadAccess( new AclKey( dstEntitySetId ) );
+                    ensureReadAccess( new AclKey( associationEntitySetId ) );
+
+                    //Ensure that we can read key properties.
+                    Set<UUID> keyPropertyTypes = dms.getEntityTypeByEntitySetId( associationEntitySetId ).getKey();
+                    keyPropertyTypes.forEach( propertyType ->
+                            accessCheck( new AclKey( associationEntitySetId, propertyType ), READ_PERMISSION ) );
+
+
+                    if ( associationEntitySetIds.putIfAbsent(
```
geekbeast: This would be so much less terrible in Kotlin.

geekbeast (approved):

src/main/java/com/openlattice/datastore/data/controllers/DataIntegrationController.java
```diff
@@ -216,7 +219,40 @@ public IntegrationResults integrateEntityAndAssociationData(
...
+        edges.forEach(
+                association -> {
+                    final var associationEntitySetId = association.getEdge().getEntitySetId();
+                    final var srcEntitySetId = association.getSrc().getEntitySetId();
+                    final var dstEntitySetId = association.getDst().getEntitySetId();
+
+                    //Ensure that we have read access to entity set metadata.
+                    ensureReadAccess( new AclKey( srcEntitySetId ) );
+                    ensureReadAccess( new AclKey( dstEntitySetId ) );
+                    ensureReadAccess( new AclKey( associationEntitySetId ) );
```
katherinebernstein: If this is called during an integration this might be duplicating these calls tens of thousands of times -- would probably be faster to collect a map of all entity sets written -> all property types written on those entity sets, and then do the access checks once

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -309,24 +310,44 @@ public Integer replaceEntityProperties(
...
-                            .getEntityTypeByEntitySetId( associationEntitySetId ).getKey();
-                    keyPropertyTypes.forEach( propertyType ->
-                            accessCheck( new AclKey( associationEntitySetId, propertyType ), READ_PERMISSION ) );
+                    final var edgeEntitySetId = association.getEdge().getEntitySetId();
+                    final var srcEntitySetId = association.getSrc().getEntitySetId();
+                    final var dstEntitySetId = association.getDst().getEntitySetId();
+
+                    entitySetIdChecks.put( new AclKey( edgeEntitySetId ), READ_PERMISSION );
+                    entitySetIdChecks.put( new AclKey( srcEntitySetId ), READ_PERMISSION );
+                    entitySetIdChecks.put( new AclKey( dstEntitySetId ), READ_PERMISSION );
```
katherinebernstein: Should these be write permissions?

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -309,24 +310,44 @@ public Integer replaceEntityProperties(
...
+                    }
                 }
         );
 
-        WriteEvent writeEvent = dgm.createAssociations( associations );
+        //Ensure that we have read access to entity set metadata.
+        accessCheck( entitySetIdChecks );
+
+        //Ensure that we can read key properties on association.
+        accessCheck( aclKeysForAccessCheck( keyPropertiesOfAssociations, READ_PERMISSION ) );
```
katherinebernstein: Do you know why this is only checking key property types? (I see this is how we were doing it before but not sure what the logic is there) (also unclear to me why it was/is checking read and not write)

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
you-leee: This call does not write any data, so I guess that's why you don't need write permission. The original checks were implemented here: https://jira.openlattice.com/browse/LATTICE-1738

geekbeast:
Should this be merged?

you-leee:
Yes

katherinebernstein:
following up on my question about checking key property types on associations?

katherinebernstein (approved):

PR #400, opened 2019-06-14
Bugfix/lattice 2048 edge audit entityset deletion

you-leee:


src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -712,21 +713,28 @@ public Integer deleteEntityProperties(
...
-        associationsEdgeKeys.forEach( edgeKey -> {
+
+        // collect association entity key ids
+        final Iterable<DataEdgeKey> associationsEdgeKeys = collectAssociations( entitySetId, entityKeyIds )
+                .stream()
+                // for soft deletes, we skip clearing edge audit entities
+                // filter out audit entity sets
+                .filter( edgeKey -> !edmService.getAuditRecordEntitySetsManager().getAuditingTypes()
+                        .getAuditingEdgeEntityTypeId()
+                        .equals( edmService.getEntitySet( edgeKey.getEdge().getEntitySetId() ).getEntityTypeId() )
```
katherinebernstein: The `edmService.getEntitySet` call is pretty heavy and will often be repeated a bunch of times since the same entity sets will likely show up in lots of DataEdgeKeys. Could probably make this a lot faster by collecting all the distinct edge entity set ids up front, loading the entity types for them once, and then filtering out the audit edge ones.

katherinebernstein (changes requested):

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
you-leee: I considered that option too, but then we have to iterate through all the associationsEdgeKeys one more time. But if the entity set load is such a heavy call, it makes more sense to collect them upfront.

katherinebernstein: I think an extra iteration will be lighter than an extra read from hazelcast for each DataEdgeKey

src/main/java/com/openlattice/datastore/data/controllers/DataController.java
```diff
@@ -712,37 +713,61 @@ public Integer deleteEntityProperties(
...
         final PostgresIterable<DataEdgeKey> associationsEdgeKeys = collectAssociations( entitySetId, entityKeyIds );
 
-        // access checks
-        final Map<UUID, Map<UUID, PropertyType>> authorizedPropertyTypes = new HashMap<>();
-        associationsEdgeKeys.forEach( edgeKey -> {
+        // collect edge entity sets
+        final var edgeEntitySetIds = associationsEdgeKeys.stream()
+                .map( edgeKey -> edgeKey.getEdge().getEntitySetId() )
+                .collect( Collectors.toSet() );
+        final var edgeEntitySets = edmService.getEntitySetsAsMap( edgeEntitySetIds );
```
katherinebernstein: It would also be lighter to use a hazelcast aggregator to filter to a set of entity set ids that match the audit edge entity type instead of loading the full entity set objects

you-leee: Yeah, right, will do!

katherinebernstein (approved):

PR #405, opened 2019-06-24
log entity set id on unauthorized delete

katherinebernstein:


you-leee (approved):
That's a time saving idea! :)

makosblade (approved):

PR #431, opened 2019-08-07
when returning organizations, filter to readable roles for user

katherinebernstein:


src/main/java/com/openlattice/controllers/OrganizationsController.java
```diff
@@ -106,15 +111,15 @@ public Organization getOrganization( @PathVariable( ID ) UUID organizationId ) {
...
-        return new Organization(
-                org.getSecurablePrincipal(),
-                org.getAutoApprovedEmails(),
-                org.getMembers(),
-                authorizedRoles,
-                org.getApps(),
-                org.getSmsEntitySetInfo(),
-                org.getPartitions() );
+
+        Set<AclKey> authorizedRoleAclKeys = authorizations
```
you-leee: If we filter roles by the current principals having read on them in `getOrganization`, shouldn't we also do it, when calling `getRoles`?

katherinebernstein: I think we do -- `getRoles` calls `getAuthorizedRoles` which loads the organization and filters roles to ones with READ permissions

you-leee: Then why not use it also here?

katherinebernstein: Since this way is more efficient since it's only making one bulk authorization call here. I could rewrite that one to make more efficient calls too though...

you-leee: I see! 👍

you-leee (approved):

makosblade (approved):

PR #432, opened 2019-08-07
make apps uninstallable

katherinebernstein:


src/main/java/com/openlattice/datastore/apps/services/AppService.java
```diff
@@ -240,6 +240,21 @@ public void installApp( UUID appId, UUID organizationId, String prefix, Principa
         organizationService.addAppToOrg( organizationId, appId );
     }
 
+    public void uninstallApp( UUID appId, UUID organizationId ) {
```
you-leee, line 243: I'm not sure about the role of the entity set created, when an app is installed, just double checking, if that needs to be handled here or not.

src/main/java/com/openlattice/datastore/apps/services/AppService.java
```diff
@@ -459,7 +474,7 @@ private void updateAppConfigsForNewAppType( UUID appId, Set<UUID> appTypeIds ) {
 
     private void ensureAppTypesAreValid( Set<UUID> appTypeIds ) {
         Set<UUID> missingAppTypes = Sets.difference( appTypeIds,
-                appTypes.keySet( Predicates.in( "__#key", appTypeIds.toArray( new UUID[] {} ) ) ) );
+                appTypes.keySet( Predicates.in( "__key", appTypeIds.toArray( new UUID[] {} ) ) ) );
```
makosblade, line 477: :here-be-dragons:

makosblade (approved):

src/main/java/com/openlattice/datastore/apps/services/AppService.java
katherinebernstein: yeah I figured it's safer to let the user delete the entity sets if they want, since it's possible they don't want to lose all their data but just don't want to use the app anymore

src/main/java/com/openlattice/datastore/apps/services/AppService.java
katherinebernstein: :slain:

PR #434, opened 2019-08-20
endpoint for bulk acl update

katherinebernstein:


src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -64,62 +60,60 @@
...
+                .filter( authorization -> !authorization.getPermissions().get( Permission.OWNER ) )
+                .map( Authorization::getAclKey )
+                .collect( Collectors.toSet() );
+
+        if ( !unauthorizedAclKeys.isEmpty() ) {
+            throw new ForbiddenException( "Only owner of securable objects " + unauthorizedAclKeys +
+                    " can access other users' access rights." );
+        }
+
+        requestsByActionType.entrySet().forEach( entry -> {
```
you-leee: You could also use Map.foreach { (key, value) -> ...}

katherinebernstein: yeah that's way cleaner :)

you-leee (approved):

makosblade (approved):
The Kotlinification must continue

geekbeast (approved):

PR #437, opened 2019-09-05
Feature/lattice 1756 data expiration

pckeyes:
controller level implementation of expiration apis and added expiration field to usage of entityset class.

### DIFF
--- a/src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
+++ b/src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
@@ -455,45 +455,14 @@ constructor(
         check(es.hasExpirationPolicy()) { "Entity set ${es.name} does not have an expiration policy" }
 
         val expirationPolicy = es.expiration
-        val sqlParams = getSqlParameters(expirationPolicy, dateTime)
-        val expirationBaseColumn = sqlParams.first
-        val formattedDateMinusTTE = sqlParams.second
-        val sqlFormat = sqlParams.third
-        return dgm.getExpiringEntitiesFromEntitySet(entitySetId, expirationBaseColumn, formattedDateMinusTTE,
-                sqlFormat, es.expiration.deleteType).toSet()
+        var expirationPT = Optional.empty<PropertyType>()
+        if (expirationPolicy.startDateProperty.isPresent) {
+            expirationPT = Optional.of(edmManager.getPropertyType(expirationPolicy.startDateProperty.get()))
+        }
+        return dgm.getExpiringEntitiesFromEntitySet(entitySetId, expirationPolicy, dateTime, es.expiration.deleteType, expirationPT).toSet()
     }
 
-    private fun getSqlParameters(expirationPolicy: DataExpiration, dateTime: OffsetDateTime): Triple<String, Any, Int> {
-        val expirationBaseColumn: String
-        val formattedDateMinusTTE: Any
-        val sqlFormat: Int
-        val dateMinusTTEAsInstant = dateTime.toInstant().minusMillis(expirationPolicy.timeToExpiration)
-        when (expirationPolicy.expirationBase) {
-            ExpirationBase.DATE_PROPERTY -> {
-                val expirationPropertyType = edmManager.getPropertyType(expirationPolicy.startDateProperty.get())
-                val columnData = Pair(expirationPropertyType.postgresIndexType, expirationPropertyType.datatype)
-                expirationBaseColumn = PostgresDataTables.getColumnDefinition(columnData.first, columnData.second).name
-                if (columnData.second == EdmPrimitiveTypeKind.Date) {
-                    formattedDateMinusTTE = OffsetDateTime.ofInstant(dateMinusTTEAsInstant, ZoneId.systemDefault()).toLocalDate()
-                    sqlFormat = Types.DATE
-                } else { //only other TypeKind for date property type is OffsetDateTime
-                    formattedDateMinusTTE = OffsetDateTime.ofInstant(dateMinusTTEAsInstant, ZoneId.systemDefault())
-                    sqlFormat = Types.TIMESTAMP_WITH_TIMEZONE
-                }
-            }
-            ExpirationBase.FIRST_WRITE -> {
-                expirationBaseColumn = "${VERSIONS.name}[array_upper(${VERSIONS.name},1)]" //gets the latest version from the versions column
-                formattedDateMinusTTE = dateMinusTTEAsInstant.toEpochMilli()
-                sqlFormat = Types.BIGINT
-            }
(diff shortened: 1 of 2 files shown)


makosblade (approved):

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
```diff
@@ -402,6 +414,87 @@ constructor(
...
+    override fun getExpiringEntitiesFromEntitySet(
+            @PathVariable(ID) entitySetId: UUID,
+            @RequestBody dateTimeAsString: String): Set<UUID> {
+        ensureReadAccess(AclKey(entitySetId))
+        val dateTime = OffsetDateTime.parse(dateTimeAsString)
+        val es = getEntitySet(entitySetId)
+        check(es.hasExpirationPolicy()) { "Entity set ${es.name} does not have an expiration policy" }
+
+        val expirationPolicy = es.expiration
+        if (expirationPolicy != null) {
```
you-leee: I think the null check is not necessary at this point, because `es.hasExpirationPolicy()` checks if expiration policy is null or not (and throws exception if it is)

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
```diff
@@ -402,6 +414,87 @@ constructor(
...
+
+    @Timed
+    @RequestMapping(
+            path = [ALL + ID_PATH + EXPIRATION_PATH],
+            method = [RequestMethod.POST]
+    )
+    override fun getExpiringEntitiesFromEntitySet(
+            @PathVariable(ID) entitySetId: UUID,
+            @RequestBody dateTimeAsString: String): Set<UUID> {
+        ensureReadAccess(AclKey(entitySetId))
```
katherinebernstein, line 452: I'm not sure this needs to be exposed via the API, since I think this will only be called internally when removing expired data, and this is a lot of work for datastore to be doing. This seems like it should go in a DataExpirationService or something in conductor-client or indexer

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
```diff
@@ -402,6 +414,87 @@ constructor(
...
+        val dateTime = OffsetDateTime.parse(dateTimeAsString)
+        val es = getEntitySet(entitySetId)
+        check(es.hasExpirationPolicy()) { "Entity set ${es.name} does not have an expiration policy" }
+
+        val expirationPolicy = es.expiration
+        if (expirationPolicy != null) {
+            val sqlParams = getSqlParameters(expirationPolicy, dateTime)
+            return dgm.getExpiringEntitiesFromEntitySet(entitySetId, sqlParams, es.expiration.deleteType).toSet()
```
you-leee: Just a minor thing, but I would put the `sqlParams` into 3 separate parameters in `getExpiringEntitiesFromEntitySet`. It's easier to understand what they stand for when reading the code

pckeyes: For sure I can do that!

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
katherinebernstein: ^nvm ignore this, I realize this is used for loading soon-to-expire data.

 I would move `getSqlParameters` out of the controller and into the service though

pckeyes: The ticket asked for the ability to review "soon to expire data", is there a way to allow this without an api?

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
pckeyes: oops! thanks for catching that

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
pckeyes: Ah okay will do

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
```diff
@@ -402,6 +414,56 @@ constructor(
...
+    override fun getExpiringEntitiesFromEntitySet(
+            @PathVariable(ID) entitySetId: UUID,
+            @RequestBody dateTimeAsString: String): Set<UUID> {
+        ensureReadAccess(AclKey(entitySetId))
+        val dateTime = OffsetDateTime.parse(dateTimeAsString)
+        val es = getEntitySet(entitySetId)
+        check(es.hasExpirationPolicy()) { "Entity set ${es.name} does not have an expiration policy" }
+
+        val expirationPolicy = es.expiration
+        var expirationPT = Optional.empty<PropertyType>()
```
makosblade: You can actually do this *very* slick `.map{}` call here instead of the `if` (if you so desire)
```
val expirationPT = expirationPolicy.startDateProperty.map {
        edmManager.getPropertyType( it )
}
```

makosblade (approved):

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
pckeyes: this is much better!

makosblade: Isn't it though!! I just found it today in reviewing this 🎉

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
```diff
@@ -402,6 +414,53 @@ constructor(
...
+            @PathVariable(ID) entitySetId: UUID,
+            @RequestBody dateTimeAsString: String): Set<UUID> {
+        ensureReadAccess(AclKey(entitySetId))
+        val dateTime = OffsetDateTime.parse(dateTimeAsString)
+        val es = getEntitySet(entitySetId)
+        check(es.hasExpirationPolicy()) { "Entity set ${es.name} does not have an expiration policy" }
+
+        val expirationPolicy = es.expiration
+        val expirationPT = expirationPolicy.startDateProperty.map { edmManager.getPropertyType(it) }
+        return dgm.getExpiringEntitiesFromEntitySet(entitySetId, expirationPolicy, dateTime, es.expiration.deleteType, expirationPT).toSet()
```
makosblade, line 459: I'd maybe also just inline `es.expiration` instead of creating the `expirationPolicy` value

katherinebernstein (approved):

PR #441, opened 2019-10-08
Feature/lattice 2228 atlas external access

pckeyes:
Controller level implementation for apis involved in atlas database access


src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -200,4 +225,11 @@ public Acl getAcl( @RequestBody AclKey aclKeys ) {
     public AuthorizationManager getAuthorizationManager() {
         return authorizations;
     }
+
+    private List<Acl> getOrganizationExternalDatabaseAcls(List<Acl> acls) {
+        return acls.stream().filter( acl -> {
+            SecurableObjectType type = securableObjectTypes.get( new AclKey( acl.getAclKey() ) );
+            return type == SecurableObjectType.OrganizationExternalDatabaseColumn || type == SecurableObjectType.OrganizationExternalDatabaseTable;
+        }).collect( Collectors.toList() );
```
makosblade: Would be better to do:
`securableObjectTypes.executeOnKeys( <allAclsPassedInAsAclKeys>, new ReadOnlyTypeCheckingAclReturningEntryProcessor() );`
That way you rely on hazelacast to exploit data locality and parallelization and cut down on total data serialized rather than having to run this filter that serially gets and serializes extra data

src/main/kotlin/com/openlattice/organizations/OrganizationExternalDatabaseController.kt
```diff
@@ -0,0 +1,197 @@
...
+    private fun ensureOwner(organizationId: UUID): AclKey {
+        val aclKey = AclKey(organizationId)
+        accessCheck(aclKey, EnumSet.of(Permission.OWNER))
+        return aclKey
+    }
+
+    private fun getExternalDatabaseObjectId(containingObjectId: UUID, name: String): UUID {
+        val fqn = FullQualifiedName(containingObjectId.toString(), name)
+        val id = aclKeysMap.get(fqn.fullQualifiedNameAsString)
+        checkState(id != null, "External database object with name $name does not exist")
```
makosblade: I feel like this should never happen but also Kotlin hates it without the nullcheck 🤔

makosblade (changes requested):

src/main/kotlin/com/openlattice/organizations/OrganizationExternalDatabaseController.kt
pckeyes: And even after checking kotlin still requires the `!!` lol. But I agree it should never happen, but if something gets broken and it does happen it's better to have the check right?

makosblade: Yeah, absolutely!

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
pckeyes: How will I return the Acl from the entry processor though? It won't have the necessary information I don't think.

you-leee: I would push down this function to `HazelcastSecurableObjectResolveTypeService` (SecurableObjectResolveTypeService interface) and that way you don't have to add `securableObjectTypes` map, only inject that class and put the function there.
Also, I agree with Drew here on creating a processor for this one. Basically an Ace consists of an AclKey + Aces, so what you can do is create a map, where the key is the AclKey and the Aces are grouped/merged together by that (do a group by AclKey). Once you have that map, you can use it to operate on the `securableObjectTypes` IMap, where the entryprocessor get's all the entries from `securableObjectTypes` which match the keys and filters the ones which have the desired type(s) to make it general. It will return the filtered map, which you can use `executePrivilegesUpdate`.

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -0,0 +1,238 @@
...
+        val aclKeys = columnIds.map { AclKey(organizationId, tableId, it) }.toSet()
+        aclKeys.forEach { aclKey ->
+            ensureOwnerAccess(aclKey)
+            authorizations.deletePermissions(aclKey)
+            securableObjectTypes.deleteSecurableObjectType(aclKey)
+        }
+        edms.deleteOrganizationExternalDatabaseColumns(organizationId, columnIds)
+    }
+
+    private fun ensureOwner(organizationId: UUID): AclKey {
```
you-leee: The return value is never used here + you could use `AuthorizingComponent.ensureOwnerAccess`

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -0,0 +1,238 @@
...
+        val aclKey = AclKey(organizationId)
+        accessCheck(aclKey, EnumSet.of(Permission.OWNER))
+        return aclKey
+    }
+
+    private fun getExternalDatabaseObjectId(containingObjectId: UUID, name: String): UUID {
+        val fqn = FullQualifiedName(containingObjectId.toString(), name)
```
you-leee: I would use rather `HazelcastAclKeyReservationService.getId(fqn)` here and that way you don't have to add the `aclKeysMap` IMap to the class.

you-leee: + you could write a function in `HazelcastAclKeyReservationService` to get ids in a batch for `deleteExternalDatabaseTables` and `deleteExternalDatabaseColumns`, so it will only tap into the hazelcast map 1 time

you-leee (changes requested):
The hazelcast map related comments

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
pckeyes: I ended up using a predicate to get the aclKeys that are mapped to the correct object types in the securableObjectTypes map, I think this will solce the underlying problem of accessing the imap repeatedly.

you-leee (approved):

src/main/java/com/openlattice/datastore/permissions/controllers/PermissionsController.java
```diff
@@ -53,6 +60,15 @@
     @Inject
     private EventBus eventBus;
 
+    @Inject
+    private ExternalDatabaseManagementService edms;
+
+    @Inject
+    private HazelcastInstance hazelcastInstance;
```
you-leee: I think this is not used anymore.

pckeyes: you're right! ty for catching that!

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -0,0 +1,229 @@
...
+    @Timed
+    @PostMapping(path = [ID_PATH + USER_ID_PATH + CONNECTION_TYPE_PATH + EXTERNAL_DATABASE])
+    override fun addHBARecord(
+            @PathVariable(ID) organizationId: UUID,
+            @PathVariable(USER_ID) userId: String,
+            @PathVariable(CONNECTION_TYPE) connectionType: String,
+            @RequestBody ipAddresses: Set<String>
+    ) {
+        ensureOwnerAccess(AclKey(organizationId))
+        if (connectionType == "local") {
```
you-leee: It's a minor thing, but I would create a ConnectionType enum for this with specifying 2 valid values and check against those.

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -0,0 +1,229 @@
...
+    @Inject
+    private lateinit var aclKeyReservations: HazelcastAclKeyReservationService
+
+    @Timed
+    @PostMapping(path = [ID_PATH + USER_ID_PATH + CONNECTION_TYPE_PATH + EXTERNAL_DATABASE])
+    override fun addHBARecord(
+            @PathVariable(ID) organizationId: UUID,
+            @PathVariable(USER_ID) userId: String,
+            @PathVariable(CONNECTION_TYPE) connectionType: String,
+            @RequestBody ipAddresses: Set<String>
```
you-leee: Are we doing some checks, if the ip addresses are valid or not? Will it break the pg_hba config if some of them are not?

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
pckeyes: So it's actually not only two types. It's just that only local connection types can't specify ip addresses. There's also host, hostssl, hostnossl that require an IP address.

pckeyes: I can still make an enum of all of those and check against each one if that would be preferable

you-leee: I think that would be cleaner, because it show exactly which arguments are excepted

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
pckeyes: good point! I think that if the IP address were invalid, when the user tried to connect it simply would not let them because the ip address they're on would not match, but we should check anyways.

pckeyes: good point! I think that if the IP address were invalid, when the user tried to connect it simply would not let them because the ip address they're on would not match, but we should check anyways.

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -0,0 +1,234 @@
...
+
+    @Inject
+    private lateinit var aclKeyReservations: HazelcastAclKeyReservationService
+
+    @Timed
+    @PostMapping(path = [ID_PATH + USER_ID_PATH + CONNECTION_TYPE_PATH + EXTERNAL_DATABASE])
+    override fun addHBARecord(
+            @PathVariable(ID) organizationId: UUID,
+            @PathVariable(USER_ID) userId: String,
+            @PathVariable(CONNECTION_TYPE) connectionType: String,
```
you-leee: You can define it as a `PostgresConnectionType` and then it's easier to check if it's a valid value

geekbeast (approved):

geekbeast (approved):

PR #457, opened 2019-12-06
Ignore empty integration requests + any instead of all

geekbeast:


src/main/java/com/openlattice/datastore/data/controllers/DataIntegrationController.java
```diff
@@ -115,6 +116,9 @@ public IntegrationResults integrateAssociations(
             @RequestBody Set<Association> associations,
             @RequestParam( value = DETAILED_RESULTS, required = false, defaultValue = "false" )
                     boolean detailedResults ) {
+        if ( associations.isEmpty() ) {
+            return new IntegrationResults( 0, 0, Optional.empty(), Optional.empty() );
```
makosblade, line 120: 🤔 
Do we have a ticket to make shuttle not upload in this case?

makosblade (approved):

you-leee (approved):

katherinebernstein (approved):

pckeyes (approved):

src/main/java/com/openlattice/datastore/data/controllers/DataIntegrationController.java
```diff
@@ -37,6 +37,7 @@
 import com.openlattice.edm.EntitySet;
 import com.openlattice.edm.set.EntitySetFlag;
 import com.openlattice.edm.type.PropertyType;
+import com.openlattice.graph.query.GraphQueryState.Option;
```
katherinebernstein, line 40: Is this import needed?

src/main/java/com/openlattice/datastore/data/controllers/DataIntegrationController.java
geekbeast: This was caused by some weirdness in case loader which calls data integration api. I double checked RestDestination in shuttle and upload is not executed if empty set of data is passed in.

src/main/java/com/openlattice/datastore/data/controllers/DataIntegrationController.java
geekbeast: I'll clear this one out in future PR.

PR #458, opened 2019-12-06
LATTICE-2330 use post for loadLinkedEntitySetBreakdown

you-leee:


makosblade (approved):
Are there API docs that need updating for this?

you-leee:
openlattice.yaml has no entry for this, if you mean that by the docs

UnsungHero97:
I implemented POST in openlattice/lattice-js#838, which is the only place that calls this API.

UnsungHero97 (approved):

katherinebernstein (approved):

PR #463, opened 2019-12-14
invalidate principals cache before syncing user (not after)

katherinebernstein:


you-leee (approved):

pckeyes (approved):

makosblade (approved):

geekbeast:
After some further though syncUser always invalidates the principals cache, so that call can be removed entirely. In addition, we should pipe whether any enrollments or organization changes were made as part of syncUser. In addition, we should also invalidate the cache when a role change happens in the organization service, so quite a bit left to do.

geekbeast (changes requested):
Just remove invalidation line entirely.

geekbeast (approved):

PR #475, opened 2020-01-25
Feature/lattice 1621 linking id obfuscation

you-leee:
- Kotlinify DataController
- Refactor SearchController entity set search calls
- Kill some more multimaps
- Clear DataIntegrationController
- Add DataIntegrationController for shuttle imports (getEntityKeyIds call)
- Clear out unused fields from controllers
- Delete unused stuff


pckeyes (approved):

makosblade (changes requested):
Will review the rest later but can you move DataController back to its original path? Moving it to a new path loses all history and makes it very hard to review :(

you-leee:
@makosblade I kotlinified DataController, that's why it's all new

makosblade:
> @makosblade I kotlinified DataController, that's why it's all new

Ohhhh okay, awesome!

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -356,40 +352,27 @@ public SearchResult executeFQNPropertyTypeSearch( @RequestBody FQNSearchTerm sea
...
+            neighbors = searchGraphService.executeEntityNeighborSearch(
+                    new EntityNeighborsFilterBulk( Map.of( entitySetId, ImmutableSet.of( entityKeyId ) ) ),
+                    principals
+            ).getOrDefault( entityKeyId, ImmutableList.of() );
         }
 
         UUID userId = spm.getCurrentUserId();
 
-        SetMultimap<UUID, UUID> neighborsByEntitySet = HashMultimap.create();
+        Map<UUID, Set<UUID>> neighborsByEntitySet = Maps.newHashMap();
```
makosblade: can use `newHashMapWithExpectedSize( 1.5 * neighbors.size )` (or some other constant >1 depending on how likely the if-statement is)

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -432,94 +415,85 @@ public SearchResult executeFQNPropertyTypeSearch( @RequestBody FQNSearchTerm sea
...
-            }
+        if ( getEntitySetsForRead( entitySetIds, principals ) != null ) {
+            result = searchGraphService.executeEntityNeighborSearch( filter, principals );
         }
 
-
         /* audit */
 
-        ListMultimap<UUID, UUID> neighborsByEntitySet = ArrayListMultimap.create();
+        Map<UUID, List<UUID>> neighborsByEntitySet = Maps.newHashMap();
```
makosblade: as above, `result.values().size * neighborsList.size` to at least not have to resize for a bit

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -555,75 +529,109 @@ public SearchResult executeFQNPropertyTypeSearch( @RequestBody FQNSearchTerm sea
...
+                                ),
+                                principals ) );
             }
         }
 
 
         /* audit */
 
-        SetMultimap<UUID, UUID> neighborsByEntitySet = HashMultimap.create();
+        Map<UUID, Set<UUID>> neighborsByEntitySet = Maps.newHashMap();
```
makosblade, line 598: another map sizing, `result.values.size() * associationMap.size() * association.size`

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
```diff
@@ -0,0 +1,922 @@
...
+            val authorizedPropertyTypesOfEntitySets = getAuthorizedPropertyTypesForEntitySetRead(
+                    entitySet, normalEntitySetIds, selectedProperties
+            )
+
+            val authorizedPropertyTypes = authorizedPropertyTypesOfEntitySets.values.first()
+            val orderedPropertyNames = LinkedHashSet<String>(authorizedPropertyTypes.size)
+
+            selectedProperties.filter { authorizedPropertyTypes.containsKey(it) }
+                    .map { authorizedPropertyTypes.getValue(it) }
+                    .map { pt -> pt.type.fullQualifiedNameAsString }
```
makosblade: I think you can combine this and the above `.map{}` ? Or will `.map{}` filter out nulls from `authorizedPropertyTypes.getValue(it)`?

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
```diff
@@ -0,0 +1,922 @@
...
+                        authorizedPropertyTypesOfEntitySets
+                )
+            } else {
+                dgm.getEntitySetData(
+                        mapOf(entitySetId to entityKeyIds),
+                        orderedPropertyNames,
+                        authorizedPropertyTypesOfEntitySets
+                )
+            }
+        } else {
```
makosblade, line 189: Can you invert this?
So it becomes:
```
if (!currentCase) {
    throw Forbidden()
}
// regular code
```

makosblade: actually I think maybe we should be checking permissions at controller endpoints, not inside of this deeper helper method but I know you just converted this so not a huge deal right now

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
```diff
@@ -0,0 +1,922 @@
...
+                requiredPropertyTypes, EdmAuthorizationHelper.WRITE_PERMISSION
+        ))
+
+        val authorizedPropertyTypesByEntitySet = authzHelper
+                .getAuthorizedPropertiesOnEntitySets(associations.keys, EdmAuthorizationHelper.WRITE_PERMISSION)
+
+        dataGraphServiceHelper.checkAssociationEntityTypes(associations)
+        val associationsCreated = dgm.createAssociations(associations, authorizedPropertyTypesByEntitySet)
+
+        val associationIds = mutableMapOf<UUID, List<UUID>>()
```
makosblade: Can size this with `associationsCreated.size()`

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
```diff
@@ -0,0 +1,922 @@
...
+
+        val currentUserId = spm.getCurrentUserId()
+
+        val entitiesCreated = mutableListOf<AuditableEvent>()
```
makosblade: Can size this one too probably

makosblade (commented):
Will take a second look at these PRs but here's a quick first pass

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -555,75 +529,109 @@ public SearchResult executeFQNPropertyTypeSearch( @RequestBody FQNSearchTerm sea
...
-                AuditEventType.LOAD_ENTITY_NEIGHBORS,
-                "Load neighbors of entities with filter through SearchApi.executeFilteredEntityNeighborSearch",
-                Optional.of( filter.getEntityKeyIds() ),
-                ImmutableMap.of( "filters", filter ),
-                OffsetDateTime.now(),
-                Optional.empty()
-        ) );
+        List<AuditableEvent> events = new ArrayList<>(
+                neighborsByEntitySet.keySet().size() + filter.getEntityKeyIds().size() );
+        UUID userId = getCurrentUserId();
```
makosblade: I think you accidentally removed the `spm.` in front of this

makosblade (commented):
Added some comments

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
you-leee: Good call! I have no idea though how likely it is, but probably 1.5 is fine with the 0.75 load factor.

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
you-leee: Yep, thanks! :)

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
you-leee: 👍

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
you-leee: yeah, if we want to consider all, we also need to take neighbors into account: `result.size() * associationMap.size() * association.size * neighbors.size()`

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
you-leee: Well, getValue would throw an error if the key is not present in `authorizedPropertyTypes`, but generally speaking `.map{}` won't filter null values.
I will combine it into one map, but I think there shouldn't be much performance difference, since both do 2 computation c1 and c2 n times.

you-leee: Btw, I just ran a test for this with 20.000.000 size list and the combined mapping was slightly faster

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
you-leee: > actually I think maybe we should be checking permissions at controller endpoints, not inside of this deeper helper method but I know you just converted this so not a huge deal right now

Yeah, but all the endpoints reading entity set data come down to this deeper function, so as long as they stay that way, I think it's nicer this way, compared to to duplicating code.


> Can you invert this?
> So it becomes:
> 
> ```
> if (!currentCase) {
>     throw Forbidden()
> }
> // regular code
> ```
Will change it to early return

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
you-leee: 👍

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
you-leee: 👍

src/main/kotlin/com/openlattice/data/controllers/DataController.kt
makosblade: oh cool! always love to have some actual proof of these things

src/main/java/com/openlattice/datastore/search/controllers/SearchController.java
```diff
@@ -555,75 +533,124 @@ public SearchResult executeFQNPropertyTypeSearch( @RequestBody FQNSearchTerm sea
...
+                var associations = associationMap.values().iterator().next();
+                if ( !associations.isEmpty() ) {
+                    expectedSize *= associations.size();
+
+                    if ( !associations.values().iterator().next().isEmpty() ) {
+                        expectedSize *= associations.values().iterator().next().size();
+                    }
+                }
+            }
+        }
```
makosblade: Okay so in this case I was just hoping to get a heuristic value for the map size. Generally, calculating the map size by iterating is way way way less performant than just leaving the map un-sized so we should probably just revert to un-sized here

you-leee: 👍

PR #476, opened 2020-02-02
Feature/lattice 2385 streamline organization dbs

geekbeast:


src/main/java/com/openlattice/controllers/OrganizationsController.java
```diff
@@ -321,8 +295,7 @@ public Void refreshDataChanges(
         // the person requesting refresh should be the owner of the organization
         ensureOwner( organizationId );
```
you-leee, line 296: If we check owner access in this call, shouldn't we also do for others/delete it from here?

src/main/java/com/openlattice/controllers/OrganizationsController.java
```diff
@@ -301,15 +283,7 @@ public OrganizationIntegrationAccount getOrganizationIntegrationAccount( @PathVa
     public Void synchronizeEdmChanges(
             @PathVariable( ID ) UUID organizationId,
             @PathVariable( SET_ID ) UUID entitySetId ) {
-        // we basically re-materialize in this case
-        final var authorizedPropertyTypesByEntitySet =
-                getAuthorizedPropertiesForMaterialization( organizationId, Set.of( entitySetId ) );
```
you-leee, line 306: Is`getAuthorizedPropertiesForMaterialization` function not deleted on purpose ?

pckeyes (approved):

dbt (approved):

you-leee (approved):

PR #479, opened 2020-02-06
Bugfix/resurrect heracles texting

katherinebernstein:


src/main/kotlin/com/openlattice/codex/controllers/CodexController.kt
```diff
@@ -81,55 +68,54 @@ constructor(
...
     }
 
-    override fun receiveIncomingText(@RequestBody message: Message) {
+    @Timed
+    @RequestMapping(path = [INCOMING + ORG_ID_PATH], method = [RequestMethod.POST], consumes = [MediaType.APPLICATION_FORM_URLENCODED_VALUE])
+    fun receiveIncomingText(@PathVariable(ORG_ID) organizationId: UUID, request: HttpServletRequest) {
 
+        val messageId = codexService.getIncomingMessageField(request, CodexConstants.Request.SID)
+        val phoneNumber = codexService.getIncomingMessageField(request, CodexConstants.Request.FROM)
+        val text = codexService.getIncomingMessageField(request, CodexConstants.Request.BODY)
```
makosblade: This series of `codexService.getIncomingMessageField(` calls is totally fine but it feels a little clunky to me to repeatedly call the same helper method rather than just adding a `codexService.processIncomingMessageFromRequest( organizationId, request, dateTime?)` that calls these and then calls `.processIncomingMessage`

src/main/kotlin/com/openlattice/codex/CodexService.kt
```diff
@@ -0,0 +1,187 @@
...
+import com.twilio.rest.api.v2010.account.Message
+import org.joda.time.DateTime
+import org.slf4j.LoggerFactory
+import java.time.Instant
+import java.time.OffsetDateTime
+import java.time.ZoneOffset.UTC
+import java.util.*
+import javax.servlet.http.HttpServletRequest
+
+class CodexService(
```
makosblade, line 40: I think this needs an `@Service`

src/main/kotlin/com/openlattice/codex/CodexService.kt
```diff
@@ -0,0 +1,187 @@
...
+        val sender = principalsManager.getUser(senderId)
+
+        /* create entities */
+
+        val contactEDK = getContactEntityDataKey(organizationId, phoneNumber)
+        val messageEDK = getMessageEntityDataKey(organizationId, dateTime, messageId, text)
+
+        /* create associations */
+
+        val associations: ListMultimap<UUID, DataEdge> = ArrayListMultimap.create()
```
makosblade, line 110: For some reason I thought we were done with `*Multimap`. Do we want to be done with `*Multimap`?

makosblade: In fact, `dgm.createAssociations` immediately just converts associations to a `Map<K, List<V>>`

makosblade (approved):

src/main/kotlin/com/openlattice/codex/CodexService.kt
katherinebernstein: ¯\_(ツ)_/¯ that's what dgm expects + what everything is currently passing it

We definitely needed to get rid of multimaps specifically for updateEntitiesInEntitySet since multimaps treat empty collection values as entry not present, and this presented problems for PartialReplace (because replacing a set of values with [] should effectively clear the existing values, but instead nothing was changed). I'm not sure what the reason for removing the other multimaps is.

makosblade: Looks like @you-leee updated her PRs to fix it in this instance already

makosblade: I also do not know the base reason but I prefer having as few types to think about as possible

src/main/kotlin/com/openlattice/codex/controllers/CodexController.kt
```diff
@@ -38,98 +31,47 @@ import javax.inject.Inject
...
-    }
+        codexService.updateMessageStatus(organizationId, messageId, status)
 
-    fun createEntitiesFromMessage(msg: Message, organizationId: UUID, appTypeId: UUID) {
-        val entities = mutableListOf<Map<UUID, Set<Any>>>()
-        val app = appService.getApp("codex")
-        val ack = AppConfigKey(app.id, organizationId, appTypeId)
-        val esid = appConfigMS.load(ack).entitySetId
+        if (status == Message.Status.FAILED || status == Message.Status.UNDELIVERED) {
+            println("Message $messageId not received or even failed to send!!! ")
```
you-leee: Why not use a logger here?

katherinebernstein: ah yeah -- that's old code that was just caught in the reformatting so I hadn't looked at it closely buuut idk why it's not using a logger

you-leee (approved):

src/main/kotlin/com/openlattice/codex/controllers/CodexController.kt
```diff
@@ -38,98 +32,51 @@ import javax.inject.Inject
...
+    @RequestMapping(path = [INCOMING + ORG_ID_PATH + STATUS], method = [RequestMethod.POST])
+    fun listenForTextStatus(@PathVariable(ORG_ID) organizationId: UUID, request: HttpServletRequest) {
 
-    fun processQueueEntry(message: Message, organizationId: UUID) {
-        createEntitiesFromMessage(message, organizationId, peopleUUID)
-        createEntitiesFromMessage(message, organizationId, contactUUID)
-        createEntitiesFromMessage(message, organizationId, messagesUUID)
-        createEntitiesFromMessage(message, organizationId, toUUID)
-        createEntitiesFromMessage(message, organizationId, fromUUID)
+        val messageId = request.getParameter(CodexConstants.Request.SID.parameter)
```
dbt, line 72: should we do a permission/ownership check here?

katherinebernstein: It's a bit tricky since all the requests will be coming from Twilio -- I can look into whether we can manually set some header for the Twilio requests that we can use for an auth check here or something

katherinebernstein: Actually we can probably do that here using the twilio session, but the other endpoint for receiving incoming messages will be tricky

src/main/kotlin/com/openlattice/codex/CodexService.kt
```diff
@@ -0,0 +1,229 @@
...
+import java.time.Instant
+import java.time.OffsetDateTime
+import java.time.ZoneOffset.UTC
+import java.util.*
+import java.util.concurrent.Executors
+import java.util.stream.Stream
+import javax.servlet.http.HttpServletRequest
+
+@Service
+class CodexService(
```
pckeyes, line 40: just wondering, why is this in datastore instead of cond-cli?

src/main/kotlin/com/openlattice/codex/controllers/CodexController.kt
katherinebernstein: nvm there appears to be a way... :D

src/main/kotlin/com/openlattice/codex/CodexService.kt
katherinebernstein: because it's only scoped to datastore right now -- not called else

pckeyes (approved):

PR #480, opened 2020-02-06
fix for get rid of listMultimap from dgm api

makosblade:


you-leee:
That api call will be changed to have the associations as a map instead of a multimap here: https://github.com/openlattice/api/pull/409

dbt (approved):

pckeyes (approved):

katherinebernstein (approved):

geekbeast (approved):

PR #488, opened 2020-03-11
LATTICE 2419 Permissions fixes for External DB tables

pckeyes:
-Api for getting all tables/columns for an org now filters on READ permission instead of ensuring OWNER access.
-Delete table now checks for OWNER access on all columns before deleting the table

### DIFF
--- a/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
+++ b/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
@@ -73,8 +73,8 @@ class DatasetController : DatasetApi, AuthorizingComponent {
     override fun getExternalDatabaseTables(
             @PathVariable(ID) organizationId: UUID): Set<OrganizationExternalDatabaseTable> {
         val tables = edms.getExternalDatabaseTables(organizationId)
-        val authorizedTableIds = getAuthorizedTableIds(tables.keys, Permission.READ)
-        return tables.filter { it.key in authorizedTableIds }.values.toSet()
+        val authorizedTableIds = getAuthorizedTableIds(tables.map { it.key }.toSet(), Permission.READ)
+        return tables.filter { it.key in authorizedTableIds }.map { it.value }.toSet()
     }
 
     @Timed
@@ -86,9 +86,9 @@ class DatasetController : DatasetApi, AuthorizingComponent {
         val columnsByAuthorizedTable = columnsByTable.filter { it.key.first in authorizedTableIds }
         return columnsByAuthorizedTable.map {
             it.key.second to it.value.filter { (columnId, _) ->
-                val authorizedColumnIds = getAuthorizedColumnIds(it.key.first, it.value.keys, Permission.READ)
+                val authorizedColumnIds = getAuthorizedColumnIds(it.key.first, it.value.map { entry -> entry.key }.toSet(), Permission.READ)
                 columnId in authorizedColumnIds
-            }.values.toSet()
+            }.map { entry -> entry.value }.toSet()
         }.toMap()
     }
 
--- a/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
+++ b/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
@@ -13,11 +13,12 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 import org.apache.olingo.commons.api.edm.FullQualifiedName
 import org.springframework.web.bind.annotation.*
 import javax.inject.Inject
-import java.util.UUID
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import com.openlattice.edm.requests.MetadataUpdate
 import org.springframework.web.bind.annotation.PatchMapping
+import java.util.*
+import java.util.stream.Collectors
 
 @SuppressFBWarnings(
         value = ["BC_BAD_CAST_TO_ABSTRACT_COLLECTION"],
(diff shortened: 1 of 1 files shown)


dbt (approved):

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -75,16 +70,20 @@ class DatasetController : DatasetApi, AuthorizingComponent {
     @GetMapping(path = [ID_PATH + EXTERNAL_DATABASE_TABLE])
     override fun getExternalDatabaseTables(
             @PathVariable(ID) organizationId: UUID): Set<OrganizationExternalDatabaseTable> {
-        ensureOwnerAccess(AclKey(organizationId))
-        return edms.getExternalDatabaseTables(organizationId)
+        return edms.getExternalDatabaseTables(organizationId).filter { isAuthorized(Permission.READ).test(AclKey(it.id)) }.toSet()
```
you-leee: If there could be many of these tables, I would rather do a bulk authorization check with `AuthorizationManager.accessChecksForPrincipals`.
For referenceyou can take a look at: `EntitySetsController.checkPermissionsForCreate` or `EntitySetsController.getPropertyMetadataForEntitySets`

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -75,16 +70,20 @@ class DatasetController : DatasetApi, AuthorizingComponent {
...
     }
 
     @Timed
     @GetMapping(path = [ID_PATH + EXTERNAL_DATABASE_TABLE + EXTERNAL_DATABASE_COLUMN])
     override fun getExternalDatabaseTablesWithColumns(
             @PathVariable(ID) organizationId: UUID): Map<OrganizationExternalDatabaseTable, Set<OrganizationExternalDatabaseColumn>> {
-        ensureOwnerAccess(AclKey(organizationId))
-        return edms.getExternalDatabaseTablesWithColumns(organizationId)
+        val authorizedColsByTable = mutableMapOf<OrganizationExternalDatabaseTable, Set<OrganizationExternalDatabaseColumn>>()
+        edms.getExternalDatabaseTablesWithColumns(organizationId).forEach { (table, columns) ->
```
you-leee: Same comment as above.

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
pckeyes: So it's not an all or nothing thing. We only want to return the tables that they do have READ on, not fail the call if they don't have READ permissions on all of them.

pckeyes: Oh nvm I see what you mean

you-leee: If you look at the return value/implementation of `accessChecksForPrincipals`, it won't throw an error, but return an `Authorization` object for each requested `AccessCheck`, which has a map `Map<Permission, Boolean> permissions` where if the value is true, the calling user had that Permission.
If you take a look at `EntitySetsController.checkPermissionsForCreate` that is doing the same thing (but for organizations) you need here (with the difference, that that function also throws an error at the end)

```
        // check read on organization
        val organizationIds = entitySets.map { it.organizationId }.toSet()

        val authorizedOrganizationIds = authorizations
                .accessChecksForPrincipals(
                        organizationIds.map { AccessCheck(AclKey(it), READ_PERMISSION) }.toSet(),
                        Principals.getCurrentPrincipals()
                )
                .filter { it.permissions.contains(Permission.READ) }
                .map { it.aclKey[0] }
                .collect(Collectors.toSet())
```
^ this code block does a bulk check for all organization ids and filters the one, where the calling user has read permission

you-leee: > Oh nvm I see what you mean

Saw this too late  :D

pckeyes: The in depth explanation is still appreciated lol

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -167,11 +171,20 @@ class DatasetController : DatasetApi, AuthorizingComponent {
...
+        val aclKeys = tableIds.map { AclKey(it) }.toSet()
         aclKeys.forEach { aclKey ->
             ensureOwnerAccess(aclKey)
         }
+        tableIds.forEach { tableId ->
+            ensureObjectCanBeDeleted(tableId)
+            val columnIds = edms.getExternalDatabaseTableWithColumns(tableId).columns.map { it.id }.toSet()
+            columnIds.forEach { ensureObjectCanBeDeleted(it) }
+            val aclKeys = columnIds.map { AclKey(tableId, it) }.toSet()
+            aclKeys.forEach { aclKey ->
```
you-leee: This can be also bulked the same way as in `getAuthorizedTableIds`, but rather then returning the authorized ones, you can compare if their size matches or not and throw a `ForbiddenException` if not

pckeyes: I'll still need `ensureObjectCanBeDeleted` though right?

you-leee: That's right

makosblade (approved):

you-leee (approved):

katherinebernstein (approved):

PR #489, opened 2020-03-11
LATTICE 2421 audit get expiring entities correctly

pckeyes:


src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
```diff
@@ -519,11 +519,25 @@ constructor(
...
 
         val expirationPolicy = es.expiration
         val expirationPT = expirationPolicy.startDateProperty.map { edmManager.getPropertyType(it) }
+
+        recordEvent(
+                AuditableEvent(
+                        spm.currentUserId,
+                        AclKey(entitySetId),
+                        AuditEventType.READ_ENTITY_SET,
+                        "EntitySetApi.getExpiringEntitiesFromEntitySet() resulted in an EntitySet READ",
```
you-leee: Not 100% sure about the audit type and message, but otherwise lgtm

you-leee (approved):

katherinebernstein (approved):

src/main/kotlin/com/openlattice/entitysets/controllers/EntitySetsController.kt
katherinebernstein: (also super minor and I realize these are all wrong in the controller but it should be EntitySetsApi instead of EntitySetApi)

pckeyes: Id have to create a new AuditEventType if the current one isn't right (which I agree it's a little off). Does it warrant its own type? I can change the message to something like "returned entity key ids" or "returned expiring entity key ids"?

you-leee: @katherinebernstein ?

you-leee: I'm not sure how we need to audit this

katherinebernstein: oh I just meant in the string on line 533

katherinebernstein: but yeah I don't really see much harm in adding a new audit type -- I guess it would prevent filtering by all updates to entity set, but that's not exactly true because you can filter on all the more fine-grained audit types that you care about. so I'd vote for adding a new audit event type.

katherinebernstein: > oh I just meant in the string on line 533

oh jk you did this change

makosblade (approved):

PR #492, opened 2020-03-17
LATTICE 2423 Fix return type, filter for columns by permission, filter data returned by read permission

pckeyes:
This PR
-Fixes the return type of the getTablesWithColumns controller
-Renames controller to clarify that returning columns returns only metadata on the columns and not any actual data
-Adds a controller to filter for returning columns that a user has a provided permission on
-Fixes controller for returning data so only data from columns with a read permission are returned

### DIFF
--- a/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
+++ b/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
@@ -79,7 +79,7 @@ class DatasetController : DatasetApi, AuthorizingComponent {
 
     @Timed
     @GetMapping(path = [ID_PATH + EXTERNAL_DATABASE_TABLE + EXTERNAL_DATABASE_COLUMN])
-    override fun getExternalDatabaseTablesWithColumns(
+    override fun getExternalDatabaseTablesWithColumnMetadata(
             @PathVariable(ID) organizationId: UUID): Set<OrganizationExternalDatabaseTableColumnsPair> {
         val columnsByTable = edms.getExternalDatabaseTablesWithColumns(organizationId)
         val authorizedTableIds = getAuthorizedTableIds(columnsByTable.keys.map { it.first }.toSet(), Permission.READ)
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -108,7 +127,13 @@ class DatasetController : DatasetApi, AuthorizingComponent {
             @PathVariable(TABLE_ID) tableId: UUID,
             @PathVariable(ROW_COUNT) rowCount: Int): Map<UUID, List<Any?>> {
         ensureReadAccess(AclKey(tableId))
-        return edms.getExternalDatabaseTableData(organizationId, tableId, rowCount)
+        val columns = edms.getExternalDatabaseTableWithColumns(tableId).columns
+        val authorizedColumnIds = getAuthorizedColumnIds(tableId, columns.map { it.id }.toSet(), Permission.READ)
+        if (authorizedColumnIds.isEmpty()) {
+            throw ForbiddenException("Missing required permissions to read table data.")
```
you-leee: Just double checking, Is this supposed to throw an exception if none of the columns are accessible or if one or more?
If it's the earlier, I would change the error message to something like "Unable to read data table $tableId. Missing ${Permission.READ} permission on all columns."

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -79,22 +79,41 @@ class DatasetController : DatasetApi, AuthorizingComponent {
 
     @Timed
     @GetMapping(path = [ID_PATH + EXTERNAL_DATABASE_TABLE + EXTERNAL_DATABASE_COLUMN])
-    override fun getExternalDatabaseTablesWithColumns(
-            @PathVariable(ID) organizationId: UUID): Map<OrganizationExternalDatabaseTable, Set<OrganizationExternalDatabaseColumn>> {
+    override fun getExternalDatabaseTablesWithColumnMetadata(
+            @PathVariable(ID) organizationId: UUID): Set<OrganizationExternalDatabaseTableColumnsPair> {
         val columnsByTable = edms.getExternalDatabaseTablesWithColumns(organizationId)
```
you-leee, line 84: This whole function body could be replaced with `getAuthorizedExternalDbTablesWithColumnMetadata(organizationId, Permission.READ)`

pckeyes: I don't think so. The point of making two apis is because the customers want to be able to see "metadata" on any tables, so we need to return the column objects even if the person asking to see them doesn't have permissions on the column objects. It specifically doesn't do a check on column object authorizations.

The `getAuthorizedExternalDbTablesWithColumnMetadata` is so one can see what permissions they have across all the tables and columns

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
pckeyes: Throws if none of the columns are accessible. Will update the message!

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
you-leee: Oh.. you're right! I totally missed the filter on columnIds.. In this case, I would create a private function:
`getAuthorizedExternalDbTablesWithColumnMetadata(organizationId: UUID, permission: Permission, filterColumnIds: Boolean = true)` and use that, since there is a lot of code in common in these 2 functions

makosblade (approved):

katherinebernstein (approved):

dbt (approved):

PR #496, opened 2020-04-01
filter principals by users and roles

jokedurnez:


jokedurnez:
@katherinebernstein we talked about this earlier, hoping there's a prettier solution.  I tried revoking permissions of the org on the columns, but that had unfortunate consequences I don't fully understand yet.  
It would be great if we can get this small fix in.
I verified it doesn't grant unexpected permissions to org-members and it didn't seem like it.

katherinebernstein:
> @katherinebernstein we talked about this earlier, hoping there's a prettier solution. I tried revoking permissions of the org on the columns, but that had unfortunate consequences I don't fully understand yet.
> It would be great if we can get this small fix in.
> I verified it doesn't grant unexpected permissions to org-members and it didn't seem like it.

so the issue is that this won't actually block anyone from reading the table objects -- since the organization principal is granted permissions on the objects, all members of the organization will inherit those permissions, so to actually prevent unauthorized reads we'd need to remove whatever logic grants the organization READ permissions

jokedurnez:
hm, I understand it is:
- table-access means having access to the metadata (all columns).  This is not by default granted to the org.
- (table,column)-access means having access to the actual data.  This is the one granted to the org, but does not propagate in data access.

This endpoint is to check which columns the users have data access to.

geekbeast:
@jokedurnez Your statement confuses me. Unless explicitly granted the org should have access to metadata, but not data. This seems backwards from what is happening and should be fixed.

We can then script or manually do the grant/revokes. If there is some pressing need, I'm happy to merge the PR, but will say it is not clear to me whether or not this will accomplish the goal It certainly will block listing, but may not block the actual read and that info may (or may not) be accessible elsewhere.

PR #498, opened 2020-04-10
change to owner access

makosblade:
Unsure if this should be Owner or Write access 🤔

### DIFF
--- a/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
+++ b/src/main/kotlin/com/openlattice/organizations/DatasetController.kt
@@ -160,8 +160,8 @@ class DatasetController : DatasetApi, AuthorizingComponent {
             @PathVariable(ID) organizationId: UUID,
             @PathVariable(TABLE_NAME) tableName: String,
             @RequestBody metadataUpdate: MetadataUpdate) {
-        ensureAdminAccess()
         val tableFqnToId = getExternalDatabaseObjectFqnToIdPair(organizationId, tableName)
+        ensureOwnerAccess(AclKey(tableFqnToId.second))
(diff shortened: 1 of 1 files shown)


katherinebernstein (approved):

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -160,8 +160,8 @@ class DatasetController : DatasetApi, AuthorizingComponent {
             @PathVariable(ID) organizationId: UUID,
             @PathVariable(TABLE_NAME) tableName: String,
             @RequestBody metadataUpdate: MetadataUpdate) {
-        ensureAdminAccess()
         val tableFqnToId = getExternalDatabaseObjectFqnToIdPair(organizationId, tableName)
+        ensureOwnerAccess(AclKey(tableFqnToId.second))
```
jokedurnez, line 164: should that be `ensureWriteAccess`?

jokedurnez:
Interesting.  Is this the same as for entitysets?  Do you need owner for table-metadata changes on entity sets?

Also, what would WRITE otherwise mean on tables?

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
makosblade: I'm not entirely sure how Write/Read/Owner map to actual postgres permissions actually but I agree that Admin is too much and that it should be either Owner or Write

geekbeast (approved):

geekbeast:
Owner makes sense to me as our rule has been that you need to have owner on a permission to grant it.

geekbeast:
Also, write would mean you could update the actual data in the table.

PR #499, opened 2020-04-10
column metadata requires table ownership

jokedurnez:


makosblade (approved):

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
```diff
@@ -176,8 +176,8 @@ class DatasetController : DatasetApi, AuthorizingComponent {
             @PathVariable(TABLE_NAME) tableName: String,
             @PathVariable(COLUMN_NAME) columnName: String,
             @RequestBody metadataUpdate: MetadataUpdate) {
-        ensureAdminAccess()
         val tableFqnToId = getExternalDatabaseObjectFqnToIdPair(organizationId, tableName)
+        ensureOwnerAccess(AclKey(tableFqnToId.second))
```
makosblade: Actually question for those who know more about these permissions. Should this be:
`AclKey(tableFqnToId.second, columnFqnToId.second)` instead?

geekbeast: This is checking ownership of the table. If you were to add the column in it would have the second argument.

geekbeast: Actually yeah, what drew said seems correct. Based on our slack discussion this should be ensureWriteAccess(...) on the two part AclKey

geekbeast (changes requested):

src/main/kotlin/com/openlattice/organizations/DatasetController.kt
katherinebernstein: FWIW updating entity set property metadata requires OWNER not WRITE (as does updating metadata of securable objects in general)

jokedurnez: ha, interesting.  What does WRITE on a table and no properties grant you as access then?

jokedurnez: Also, wrt Drew's comment: 
granting write on a AclKey(table,column) allows you to change the data, not the metadata.  There should be a separate level.  Our summer intern works on metadata and should not be allowed to change the data.
(if there would be work on a separate level, can we please merge this because it's blocking)

makosblade: > Also, wrt Drew's comment:
> granting write on a AclKey(table,column) allows you to change the data, not the metadata. There should be a separate level. Our summer intern works on metadata and should not be allowed to change the data.
> (if there would be work on a separate level, can we please merge this because it's blocking)

Intuition says that the level you're describing is `OWNER` 🤔

makosblade (approved):

geekbeast:
Are we doing access per column? This looks like you can modify metadata for any column if you have write on the table metadata. This is probably fine, we should check what we do with entity sets and make sure is consistent (i.e what access check we do when update it).

jokedurnez:
Yeah, correct.  Right now having access on the table means having access to metadata for all columns (like with entity sets).

Giving READ on AclKey(table,column) grants `select` on that column.

geekbeast (approved):

PR #508, opened 2020-04-20
NPE on search

katherinebernstein:


makosblade (approved):

src/main/java/com/openlattice/datastore/services/DatastoreElasticsearchImpl.java
```diff
@@ -1508,12 +1508,13 @@ public void verifyRunner() throws UnknownHostException {
             UUID entitySetId,
             Map<UUID, DelegatedUUIDSet> authorizedPropertyTypesByEntitySet ) {
         Map<UUID, Map<String, Float>> fieldsMap = Maps.newHashMap();
-        authorizedPropertyTypesByEntitySet.get( entitySetId ).forEach( propertyTypeId -> {
-            String fieldName = getFieldName( propertyTypeId );
+        authorizedPropertyTypesByEntitySet.getOrDefault( entitySetId, DelegatedUUIDSet.wrap( ImmutableSet.of() ) )
```
geekbeast, line 1511: Do you need to wrap?

geekbeast (commented):
Quick question.

src/main/java/com/openlattice/datastore/services/DatastoreElasticsearchImpl.java
katherinebernstein: I didn't. see a. way to. do like DelegatedUUIDSet.empty() or something like that :/

geekbeast: I'm pretty sure that putting Set<UUID> would be fine, unless this gets serialized into an IMap at some point.

geekbeast (approved):
It should be fine for now, pretty minor.

PR #510, opened 2020-04-24
support. for sending codex media attachments

katherinebernstein:


src/main/kotlin/com/openlattice/codex/CodexService.kt
```diff
@@ -75,8 +76,10 @@ class CodexService(
     val twilioQueue = HazelcastQueue.TWILIO.getQueue(hazelcast)
     val feedsQueue = HazelcastQueue.TWILIO_FEED.getQueue(hazelcast)
 
+    val BASE_PATH = "https://api.openlattice.com"
```
makosblade: 🤔

geekbeast: +1 it this for setting up the call back hook?

geekbeast (commented):
Quick question.

src/main/kotlin/com/openlattice/codex/CodexService.kt
katherinebernstein: Yeah -- this is how we've been doing it (I just moved it into a constant in this PR) but if you have a better way to do it that would be fantastic

katherinebernstein: I guess we could add it to twilio.yaml :/

katherinebernstein: https://github.com/openlattice/conductor-client/pull/815

src/main/resources/twilio.yaml
```diff
@@ -1,3 +1,4 @@
-sid: 'nope'
-token: 'slope'
-shortCode: ""
\ No newline at end of file
+sid: "nope"
+token: "slope"
+shortCode: ""
+callbackBaseUrl: "https://api.openlattice.com"
```
makosblade: should this be localhost? b/c we use local file for local testing?

makosblade (approved):

src/main/resources/twilio.yaml
katherinebernstein: It can't be localhost since twilio will use this on their end to send messages over the network, so I usually use ngrok to test locally. I can just leave it blank though to avoid confusion

makosblade: Haaaaa, duuuhh 🤦 ty

geekbeast (approved):

PR #522, opened 2020-06-13
Update usage of getLastAclKeySafely

geekbeast:


src/main/java/com/openlattice/datastore/apps/services/AppService.java
```diff
@@ -74,6 +75,7 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.springframework.stereotype.Service;
```
makosblade: 🤔

makosblade (approved):

src/main/java/com/openlattice/datastore/apps/services/AppService.java
geekbeast: Should be fixed was a left over import.

PR #523, opened 2020-06-13
maybe perf idk

katherinebernstein:
in retrospect, definitely perf

### DIFF
--- a/src/main/java/com/openlattice/datastore/apps/services/AppService.java
+++ b/src/main/java/com/openlattice/datastore/apps/services/AppService.java
@@ -60,7 +60,9 @@ import com.openlattice.organizations.HazelcastOrganizationService;
 import com.openlattice.organizations.Organization;
 import com.openlattice.organizations.roles.SecurePrincipalsManager;
 import com.openlattice.postgres.mapstores.AppConfigMapstore;
+
 import java.util.HashSet;
+
 import org.apache.olingo.commons.api.edm.FullQualifiedName;
 
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


geekbeast (approved):

src/main/java/com/openlattice/datastore/apps/services/AppService.java
```diff
@@ -298,28 +298,38 @@ public AppType getAppType( FullQualifiedName fqn ) {
         List<AppConfig> availableConfigs = Lists.newArrayList();
         App app = apps.get( appId );
 
+        Map<UUID, String> appTypeFQNSById = Maps.newHashMapWithExpectedSize( app.getAppTypeIds().size() );
+        appTypes.getAll( app.getAppTypeIds() ).values().forEach( appType ->
```
makosblade, line 302: could use an EP instead so you don't have to serialize all of the keys as part of the return data

src/main/java/com/openlattice/datastore/apps/services/AppService.java
```diff
@@ -298,28 +298,38 @@ public AppType getAppType( FullQualifiedName fqn ) {
...
+                appTypeFQNSById.put( appType.getId(), appType.getType().getFullQualifiedNameAsString() )
+        );
+
         Map<UUID, Organization> orgsById = StreamUtil.stream( organizations )
                 .collect( Collectors.toMap( Organization::getId, Function.identity() ) );
+        int numKeys = orgsById.size() * app.getAppTypeIds().size();
+
+        Set<AppConfigKey> keysToLoad = Sets.newHashSetWithExpectedSize( numKeys );
+        orgsById.keySet().forEach( organizationId ->
+                app.getAppTypeIds().forEach( appTypeId ->
```
makosblade, line 312: should probably store `app.getAppTypeIds()` as a variable (tho it'll likely be optimized away anyway)

src/main/java/com/openlattice/datastore/apps/services/AppService.java
katherinebernstein: trufth

PR #528, opened 2020-06-26
Remove unused code + fix warnings

katherinebernstein:


makosblade (approved):

src/main/java/com/openlattice/controllers/OrganizationsController.java
```diff
@@ -125,10 +95,9 @@
                         .map( AuthorizationUtilsKt::getLastAclKeySafely )
         );
 
-        return StreamSupport.stream( orgs.spliterator(), false ).map( org -> {
-            org.getRoles().removeIf( role -> !authorizedRoles.contains( role.getAclKey() ) );
-            return org;
-        } ).collect( Collectors.toList() );
+        return StreamSupport.stream( orgs.spliterator(), false ).peek( org ->
```
UnsungHero97, line 98: Any reason not to use `StreamUtil.stream`, since it does the `spliterator()` and `false` already?

UnsungHero97, line 98: `map()` should work just fine, so why use `peek()`? The docs say:

> This method exists mainly to support debugging, where you want to see the elements as they flow past a certain point in a pipeline:

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#peek(java.util.function.Consumer)

src/main/java/com/openlattice/datastore/directory/controllers/PrincipalDirectoryController.java
```diff
@@ -20,24 +20,12 @@
...
-import com.openlattice.authorization.AclKey;
-import com.openlattice.authorization.AuthorizationManager;
-import com.openlattice.authorization.AuthorizingComponent;
-import com.openlattice.authorization.DbCredentialService;
-import com.openlattice.authorization.Permission;
-import com.openlattice.authorization.Principal;
-import com.openlattice.authorization.PrincipalType;
-import com.openlattice.authorization.Principals;
-import com.openlattice.authorization.SecurablePrincipal;
+import com.openlattice.authorization.*;
```
UnsungHero97, line 28: Why do you expand `import package.thing.*` in some places, but collapse to it in others? For example, in this file, you go from:

```
import com.openlattice.authorization.AuthorizationManager;	
import com.openlattice.authorization.AuthorizingComponent;	
import com.openlattice.authorization.DbCredentialService;	
import com.openlattice.authorization.Permission;	
import com.openlattice.authorization.Principal;	
import com.openlattice.authorization.PrincipalType;	
import com.openlattice.authorization.Principals;	
import com.openlattice.authorization.SecurablePrincipal;
```
to:
```
import com.openlattice.authorization.*;
```

In another file, you go from:
```
import java.util.*;
```
to
```
import java.util.Collection;
import java.util.LinkedHashSet;
import java.util.List;
import java.util.Map;
import java.util.Set;
import java.util.UUID;
```

src/main/java/com/openlattice/controllers/OrganizationsController.java
katherinebernstein: Not that I know of, but I wasn't sure why it was written that way in the first place so I left it as is

katherinebernstein: to fix the compiler warnings (since we're not returning a different object that's being mapped, we're only modifying it, the compiler gets angry)

src/main/java/com/openlattice/datastore/directory/controllers/PrincipalDirectoryController.java
katherinebernstein: my intellij import settings are 7 non-static / 10 static

geekbeast (approved):

PR #538, opened 2020-08-12
filter out null bytes from codex messages

katherinebernstein:


makosblade (approved):
:'(

geekbeast (approved):
My one concern here is that we don't understand why this is happening. Have we reached out to Twilio?

makosblade:
> My one concern here is that we don't understand why this is happening. Have we reached out to Twilio?

It sounded like the issue is that Twilio just passes on the source encoding, in this case GSM-7 which allows null bytes that get v.sad when you attempt to interpret them as UTF-8

geekbeast:
Hmm, why not? http://opensmpp.org/

PR #569, opened 2020-10-23
Feature/optionally exclude property types from entity set metadata search

katherinebernstein:


src/main/java/com/openlattice/datastore/services/DatastoreElasticsearchImpl.java
```diff
@@ -1252,7 +1253,14 @@ public SearchResult executeEntitySetMetadataSearch(
                 .actionGet();
 
         List<Map<String, Object>> hits = Lists.newArrayList();
-        response.getHits().forEach( hit -> hits.add( hit.getSourceAsMap() ) );
+        response.getHits().forEach( hit -> {
```
makosblade, line 1256: Isn't this file deprecated in favor of the Kotlin version? Can we remove this class/file yet?

makosblade (approved):

src/main/java/com/openlattice/datastore/services/DatastoreElasticsearchImpl.java
katherinebernstein: This is actually still the one that's being used instead of the kotlin one :( I was tempted to remove it and switch over but I figured that can be a separate PR since we should probably do some extra testing there to make sure that switchover goes smoothly

makosblade: 😢

geekbeast (approved):

